PR #70, opened 2015-01-08
Added CDI implementations

danielsoro:


akullpp:
Weld is the RI but I really don't know what's awesome about it... Dagger is interesting to learn and does some things very well and in a unique way for comparison.

PR #72, opened 2015-01-11
add JGraphT

yfpeng:
add JGraphT, a free Java graph library that provides mathematical graph-theory objects and algorithms.


akullpp:
I added it with some doubts. We had a scientific category but the state of scientific libraries in Java is unfortuntely, very bad.

PR #73, opened 2015-01-13
add Fastjson

edward9145:
Fastjson is a JSON processor (JSON parser + JSON generator) written in Java:


akullpp:
I decided not to add it for the time being. However the issue can be reopened.

The issue seems that the entire target audience is Chinese, documentation is Chinese, all discussion is in Chinese.

PR #74, opened 2015-01-15
add two JAVA image library

ThomasLau:
these two Java image scaling library(Thumbnailator,Imgscalr) are both simple and efficient   image-scaling image-scaling library implemented in pure Java，and with no third library dependent.


ThomasLau:
cool,thanks

PR #80, opened 2015-03-06
Add Feign to REST Frameworks

jon-ruckwood:
Added Netflix's Feign library which aims to ease the building of HTTP clients.


akullpp:
Yes it is indeed. So your suggestion is to remove Feign?

dominicfarr:
Yes, remove Feign. Netflix has great solutions that stand on their own.

I would go with the original and keep the community focus behind that open source project, not the imitator. 

Not sure how @JakeWharton feels about Netflix's Feign?

JakeWharton:
I know the person who wrote it. Apparently there were requirements which could not be met with Retrofit internally and so they wrote a very similar approach but tailored to their use case.

dominicfarr:
@JakeWharton, A pull request would make sense in that case. 
Anyway, Netflix were very clear about its origins; and everyone seems happy with the situation. 
Its inclusion in the list stands with you @akullpp

PR #82, opened 2015-03-09
Added XML processing section with VTD-XML library

AlphaHinex:
Added XML section and VTD-XML library which is well known by its performance.
And also did some clean work.


akullpp:
It's quite old and there are no updates as I know. Also you should have more than one item to start a list.

PR #84, opened 2015-03-14
Embedded key-value storage libraries

krisskross:


akullpp:
My issue is that I don't want stuff that just adds a Java API to something (awesome) Not-Java. That isn't what this list is about.

Ports are another critical issue, if they add more value then it's fine. If it's just another k/v store then not really awesome.

PR #87, opened 2015-03-14
added Distributed Databases

krisskross:


akullpp:
I'm not sure about this but we will see how it goes.

PR #115, opened 2015-04-21
Setting back restx ?

fcamblor:
[Restx](http://restx.io) was initially integrated with PR #29 from @simonbasle and merged by @akullpp

This addition was rolled back in adb74f4 without real justification in commit message by @akullpp :(

I think `restx` (author : @xhanin) follows what is described in `contributing.md` as it is (as far as I know) absolutely unique in its approach and function (don't know any rest framework based on annotation processor)

Plus it as some features I didn't see elsewhere :
- Annotation processing everywhere (=no reflection => great perfs)
- Type safe DI (we have others like dagger, but I don't know of fully integrated in a web framework)
- API docs far richer than swagger ones
- Specs-oriented endpoint testing (example [here](https://github.com/restx/restx/blob/master/restx-server-testing/src/test/resources/specs/sessions/should_disconnection_be_successful.spec.yaml))


akullpp:
Good catch. Won't happen again, all removals require an issue with justification from now on.

PR #117, opened 2015-04-21
Added HK2 to Dependency Injection section

facundofarias:


akullpp:
Would be interesing to know why one should favor it and not Guice.

PR #119, opened 2015-04-21
Added Functional Java

danieldietrich:
Created a new section 'Functional Programming' containing one popular entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ A curated list of awesome Java frameworks, libraries and software. Inspired by o
 *Libraries that facilitate functional programming in Java.*
 
 * [Functional Java](http://www.functionaljava.org) - Implements numerous basic and advanced programming abstractions that assist composition oriented development.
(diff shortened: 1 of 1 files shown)


akullpp:
Wouldn't it be better suited to add it to the Misc section since it's only one entry? Or you could add other functional libraries.

danieldietrich:
Yes. I think it's worth to open a section dedicated to FP as of the introduction of lambdas with Java 8.

There are two functional libraries which gained popularity within the last year:

* [jOOL](https://github.com/jOOQ/jOOL)
* [Javaslang](https://github.com/javaslang/javaslang) / [website](http://javaslang.com)

The latter is my own project, so an upvote of someone else would be great in order to not let it look like self-promotion.

akullpp:
Please add them as well.

danieldietrich:
Thank you, Sir! :-)

ggalmazor:
+1

luisartola:
+1

krisskross:
+1

PR #125, opened 2015-04-21
parquet

krisskross:


akullpp:
I'm trying to think about what Parquet is close to. I've never used it, just read about it. Is it really a whole new section? Or is it a tool storage systems, a kind-of serialization?

krisskross:
I would say that Parquet is mainly a columnar file format (as opposed to row file format like csv files), but it also include code for encoding and decoding it. Most of this is orthogonal to Hadoop even though Parquet was invented from the perspective of web scale datasets. So the main use case is probably Hadoop but you can easily use it as a jar library and encode and read Parquet files without Hadoop.

Now when I think of it, there is actually another columnar file format called ORC but AFAIK it's not distilled into a reusable library and doesn't have an authoritative website.

PR #127, opened 2015-04-22
Add native section with JNA

jvmccarthy:
[JNA](https://github.com/twall/jna) (Java Native Access) is a popular library for interacting with native libraries.  There are other libraries for dealing with native libraries ([BridJ](https://github.com/nativelibs4java/BridJ), [JACOB](http://sourceforge.net/projects/jacob-project/) for Windows COM, [JNAerator](https://github.com/nativelibs4java/JNAerator)), so I feel like native is worthy of a new section.  However, we can also add JNA to Miscellaneous if you like.  
Thanks for maintaining this list.


akullpp:
The Misc section becomes to filled, I think it is better to create a new section although it is only one library.

PR #131, opened 2015-04-22
Adding section for Cluster Management (Frameworks)

wsorenson:
Both Apache Aurora and Singularity are notable frameworks in-use by large web companies running thousands of services and are written in Java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ A curated list of awesome Java frameworks, libraries and software. Inspired by o
 - [Awesome Java](#awesome-java)
     - [Build Tool](#build-tool)
     - [Bytecode Manipulation](#bytecode-manipulation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -152,6 +153,13 @@ A curated list of awesome Java frameworks, libraries and software. Inspired by o
 * [Apache HBase](http://hbase.apache.org) - Apache HBase is the Hadoop database, a distributed, scalable, big data store.
 * [Infinispan](http://infinispan.org/) -  Distributed and highly concurrent key/value datastore used for caching.
 
+## Distributed Deployment
```
akullpp: Is it really only deployment? I would say it's kind of cluster management software, isn't it?

wsorenson: Yes it is - I was struggling with the right name.

PR #135, opened 2015-04-24
Add a section for CSV

jbax:
Adding [uniVocity-parsers](https://github.com/uniVocity/univocity-parsers) to the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ A curated list of awesome Java frameworks, libraries and software. Inspired by o
     - [Code Analysis](#code-analysis)
     - [Compiler-compiler](#compiler-compiler)
     - [Continuous Integration](#continuous-integration)
+    - [CSV](#csv)
     - [Database](#database)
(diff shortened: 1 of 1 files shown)


akullpp:
Disclosure? You are the author of the said framework with about 60 stars.

jbax:
Yes, I'm the author. This is a new library but it's quickly picking up steam because it is very superior than what other libraries offer, both in terms of speed and out-of-the box functionalities. Apache camel already introduced integration this parser. See in http://camel.apache.org/univocity-parsers-formats.html

akullpp:
Thanks for clarification. But if it is superior, why create a whole new section since adding other CSV libraries won't be necessary?

Would it be acceptable to add the library to the Datastructures section and change the purpose of the section to read/write specific datastructures like CSV?

jbax:
CSV is not exactly a data structure, but a text format such as JSON, which also has a specific section. I created the CSV section to be consistent.

PR #141, opened 2015-04-28
Added LoganSquare in JSON libraries

4k3R:
[Logan Square](https://www.youtube.com/watch?v=HUh1PJYE8KA&index=134&list=WL), according to their stats, outperforms GSON's and Jackson's databind library by 400%. I think it's a valuable addition, and it's gaining wide popularity! Supported on Android and Java!


akullpp:
I don't know about the video but I will add it. However, it would be interesting to know why it is faster than Jackson if it is based on (a part of) Jackson.

4k3R:
Yikes, link changed! Extremely sorry.. Don't worry about the merged commit. Link for Logan Square is correct in that!

PR #144, opened 2015-04-29
Added Genson to JSON libraries

acisternino:
Genson is a powerful and easy to use Java and Scala to JSON conversion library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,7 @@ A curated list of awesome Java frameworks, libraries and software. Inspired by o
 
 *Libraries that simplify JSON processing.*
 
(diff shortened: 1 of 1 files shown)


akullpp:
Why use it over the other libraries since it doesn't seem to be adopted very well.

acisternino:
Well, it has some peculiar characteristics that makes it at least as interesting as GSON.
I could repeat what is explained in its home page but I believe that link makes a very good job explaining why, at only 300KB without any dependency, it looks at least as compelling as Jackson speed-wise and as GSON feature-wise.

acisternino:
Thanks Andreas!

PR #147, opened 2015-04-30
Add gatling load testing tool

atomfrede:


akullpp:
Does it have a Java DSL and/or API? We don't add enduser applications that are just written in Java.

atomfrede:
Gatling is a Scala DSL to define load testing scenarios that can be easily versioned and run during ci build (via maven, ant, gradle...). So it's an alternative to jMeter

PR #148, opened 2015-05-01
Added ADT4J - Algebraic Data Types for Java

jbgi:
Quite unique and useful library!


talios:
Big fan of this library. Has served me well for cases I want/need ADT style processing in java. +1 would declare `<dependency>` again!

Issue #94, opened 2015-03-25
Big Cleanup


akullpp:
No offense meant, just discussing ideas.

# Delete:

## Cloud Continuous Integration

Reason for removal:
1. Adaption rate really low
2. Extremely high fluctuation
3. Money grabbers

* CircleCI
* Codeship
* Travis (only one I'm considering not to delete)
## Distributed Applications

* JGroups, kind of unique application scope but outdated

## Machine Learning

* Apache Hadoop - while still in use, the map/reduce paradigm has strongly fallen from grace with leightweight approaches like Apache Spark provides, Mahout has switched also I think.
* Apache Mahout - the switch from map/reduce has to be evaluated.

## REST Frameworks

* Feign - Copy of Retrofit with less adaption rate. Although I currently see application of Feign in a Spring Boot project.

## Templating

* Velocity - Kind of defies what this list is about, last update was 2010 and there are alternatives.

# Merge

* **Database** (7) and **Distributed Databases** (2) or split into Databases and Database Tools

# Thoughts

Projects hosted on Google Code should be checked for an official GitHub repository. If there's none, it could be a strong indicator that it is dead. These are:

* [Thumbnailator](https://code.google.com/p/thumbnailator) - Thumbnailator is a high-quality thumbnail generation library for Java.
* [Lanterna](https://code.google.com/p/lanterna/) - Easy console text GUI library similar to curses.

rasheedamir:
@akullpp why would you like to delete "Liquibase"? That is the most commonly used database migration tool. Maybe I have misunderstood the purpose of the list of stuff we are adding in "awesome-java"?

akullpp:
Thanks for the comment which brings up one of the most fundamental issues which was discussed early on and needs to be updated in the contribution guidelines too.

Regarding "most commonly used": Take for example Apache Ant, I would argue that Ant is probably more widespread than Gradle and there are fields of application where Ant is in fact more suited than Gradle.

But you have a situation where a colleague is complaining about how bloaty Java is and you should do everything in JavaScript or Golang because they are awesome. What do you show them to convince them that Java is awesome? Your 10k lines Ant XML-configuration which evolved over 20 years and is the holy grail of configuration or the new shiny Gradle which admittedly gives you sometimes headaches?

The main point here is the adjective "awesome" which many "awesome lists" do not define or care about and just list every damn plugin because it's easy to do so. You neither have to think nor judge nor re-evaluate the decisions.

But I see your point and would like more input.

[...]

PR #150, opened 2015-05-07
added cglib

bin5:


akullpp:
Won't be added here, see git log.

But we will add it in the Ancients section, see #94.

PR #152, opened 2015-05-07
added Eclipse Hudson

mosabua:


akullpp:
Won't be added, see one of the discussions about Hudson.

We can add it to the Ancients section see #94.

mosabua:
This would be completely wrong. Hudson is a separately maintained powerful CI server from Jenkins that has regular releases and among other things builds all projects at Eclipse. It also builds all Oracle weblogic software and many others.  Feature wise it is quite different from jenkins with things like config inheritance and a rich security model out of the box in the open source version. I think leaving it out is wrong.

akullpp:
I see both sides of the discussion. Hudson is used in very large projects as is Ant. Hudson is actively maintained. However, outside of the Eclipse community it is a bit frowned upon and often associated with outdated technologies. I don't know why that is.

Hudson was added and removed and added and removed. So I don't know what the right decision is. Personally, I wouldn't recommend it over Jenkins but this is only because I've used Hudson in projects with an old stack and Jenkins always with a newer stack.

mosabua:
Imho for completeness I would have both. They are very powerful and mature CI servers with their pros and cons, but are probably better than some of the other ones you have listed (without naming names...).

PR #166, opened 2015-05-26
add tweetnacl.java security library

sequoiar:
rewrite tweetnacl.c in pure Java


akullpp:
It's a rewrite, you are the author, it is not widely used, it doesn't fill a niche.

PR #195, opened 2015-08-04
Add Fugue: Functional extensions (Guava, Java 8)

jbgi:
Nice, lightweight, well-tested, functional programming library to complement Guava (2.x serie) or Java 8 (3.x serie).
In particular it provides Option and Either types similar to the Scala library as well as a Pair.
By Atlassian and contributors.


alexwei:
+1

PR #203, opened 2015-08-13
Adding CDI in the dependency injection section

agoncal:


akullpp:
I don't see a benefit adding a specification.

agoncal:
CDI is the specification, but Weld and OpenWebBeans are both implementations

PR #222, opened 2015-09-09
added new sites to read.

omerhakanbilici:
Vogella and Mkyong java blogs added.


akullpp:
They are tutorial sites with very basic (not even fudamental) Java. Also their quality is not something I want to endorse.

PR #223, opened 2015-09-09
Added the byteman library 

geoand:
Byteman is a bytecode manipulation lirbary but is distinct in purpose from the tools already mentioned, basically because it's used for injecting faults during testing or troubleshooting in production


akullpp:
Thanks for mentioning the distinct points

geoand:
You're welcome!

PR #229, opened 2015-09-20
Put Liquibase back under database

levonk:
Liquibase is a far more architecturally sound database change management than Flyway.   Liquibase abstracts underlying databases by not using SQL for most of the change scripts.   It also offers database refactorings (like rename column) that could be modified to do whatever is most efficient for a particular database with using proprietary SQL.


akullpp:
Thanks for the clarification.

PR #241, opened 2015-10-10
Added Stanford CS 106A to "Sites to read" section

virustotalop:


akullpp:
Looks like a more general class and not Java-centric. This list is Java-centric and should only mention books and websites that are focused on Java.

PR #244, opened 2015-10-20
Add Derive4J

jbgi:
Derive4J: first class support for sum types (structural pattern matching),
laziness and more: https://github.com/derive4j/derive4j


akullpp:
See #246

PR #246, opened 2015-10-21
Add the derive4j project

573:
Add the derive4j project to the functional programming section.


akullpp:
Although the project is quite new, I will add it since it is recommendable that new functional libraries are developed.

PR #249, opened 2015-10-29
Added Spring REST

andreybleme:
Added a nice Spring module to easily build RESTful web services using Spring annotations.

If you don't think it's appropriate, feel free to disregard the pull request. But would be awesome to let people know about this awesome Spring module.

**:)**


akullpp:
We have Spring as top-level project and Spring Boot. That should be sufficient.

Issue #261, opened 2015-11-19
FlexyPool


vladmihalcea:
Although it's a framework I wrote myself, I think it's worth taking a look on it since it might fit in the "(d) a niche product that fills a gap" section.

[FlexyPool](https://github.com/vladmihalcea/flexy-pool) brings all sorts of connection-related metrics to the most common connection pooling solutions. It also allows you to define fail-over strategies, to reduce the likelihood of getting a connection acquisition failure.

akullpp:
Looks great, please open a PR.

PR #263, opened 2015-11-20
#261 Add FlexyPool to the Database section

vladmihalcea:
#261 Add FlexyPool to the Database section


vladmihalcea:
I saw the mistake and created a new branch to fix it, but somehow I managed to send you the first one with the extra parenthesis. This one should be fine.

PR #282, opened 2015-12-14
Update README.md

raydac:
I offer to add the JBBP (Java Binary Block Parser) link into the list, it can be useful for many java developers who work with binary data


akullpp:
Not only the description is error-ridden with spelling issues but also the documentation of the project itself. Please fix it first.

PR #292, opened 2015-12-30
add tadpole db hub

KimKyung-man:
Add tadpole db.


akullpp:
How is it related to Java as a programming language?

KimKyung-man:
Tadpole db is a software developed in java

PR #312, opened 2016-01-25
Travis Update 

ReadmeCritic:


ReadmeCritic:
Hm looks like there was a temporary issue with projectreactor.io, closing

PR #314, opened 2016-01-30
Adding commercial reference to commercial tool

levonk:
Fixes two issues:
- The alphabetical ordering within a section should be preserved
- Properly identifies the tool as a commercial (free for open source) tool


akullpp:
Checker is already included.

PR #315, opened 2016-01-30
Adding Checker Framework

levonk:
Adding powerful static analysis tool


pron:
Checker is already included.

PR #316, opened 2016-01-30
Adding Code Conventions section

levonk:
Would be helpful to see a list of various coding conventions.   Adding Google's style guide for now.


akullpp:
Already added, doesn't need a section unless there are more influential styleguides

PR #321, opened 2016-02-23
Added Boxfuse

axelfontaine:
I've added Boxfuse as it is both highly innovative and unique in its approach to deploy Java applications to AWS. It fuses JVM applications into machine images that can be deployed locally on VirtualBox and in the cloud on AWS, giving complete environment parity. It comes with out of the box integrations for Spring Boot, Dropwizard, Grails, Play, Tomcat and TomEE applications. All deployments follow the principles of Immutable Infrastructure where servers are replaced, not updated.

_Disclaimer: I am the creator of Flyway and the founder and CEO of Boxfuse_

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ A curated list of awesome Java frameworks, libraries and software.
     - [Database](#database)
     - [Date and Time](#date-and-time)
     - [Dependency Injection](#dependency-injection)
+    - [Deployment](#deployment)
     - [Development](#development)
     - [Distributed Applications](#distributed-applications)
     - [Distributed Databases](#distributed-databases)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,6 +239,12 @@ A curated list of awesome Java frameworks, libraries and software.
 * [Guice](https://github.com/google/guice) - Lightweight but powerful framework that completes Dagger.
 * [HK2](https://hk2.java.net) - Light-weight and dynamic dependency injection framework.
 
+## Deployment
```
akullpp: Is a new category with a single entry really necessary? There is a `Distribution` section which might be suited.

axelfontaine:
Fair enough. I now removed the new deployment category and added it to distribution.

PR #328, opened 2016-03-11
Add beads audio library

alectic:


akullpp:
No new section for a single library, please use Misc or something else.

alectic:
The rules are to open a single pull request for every add, so I couldn't have added more than one in one PR. I would have liked to also add JSyn to the section.

PR #330, opened 2016-03-17
Add mybatis-page-helper

mallim:
Mybatis-page-helper offers pagination to mybatis


akullpp:
We can't add a library where the documentation is Chinese-only, this would restrict the group of users who can actually use it.

Issue #337, opened 2016-03-27
Proposal: jabba - Java Version Manager inspired by nvm


shyiko:
Hi. 

Would you consider adding [jabba](https://github.com/shyiko/jabba) (Java Version Manager inspired by nvm/gvm/rvm) to this list? It's fairly new so I understand why it might not get be accepted but it's quite unique in Java world in that it's the only (to my knowledge) tool that allows automated Oracle JDK/Zulu OpenJDK installation & seamless switching between different releases.

Thank you.

(disclosure: I'm the author of the project)

akullpp:
Sounds exactly like the thing we seek, please do a PR. :)

PR #338, opened 2016-03-27
Added jabba

shyiko:
#337. 

Thank you!


sgwerder:
I see why you couldn't call it "jvm" ;D

PR #339, opened 2016-03-30
Added new entry JavaCodeGeeks to Websites section

vidyuthd:


akullpp:
We deleted it because it's just bloat and advertisement.

PR #341, opened 2016-04-01
Adds networking library urnlib

claussni:
This library implements the a class for representing an Uniform Resource Name (URN). It implements parsing and serialization according to the constraints defined in [RFC 2141](https://tools.ietf.org/html/rfc2141). It's released on [Maven Central](http://search.maven.org/#artifactdetails|de.slub-dresden|urnlib|1.0.0|jar) and aims to serve as a network identifier handling library just like the native Java URL and URI classes.

I'm not entirely sure `Networking` is the proper category, but I also don't see any other category it might fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [Nifty](https://github.com/facebook/nifty) - Implementation of Thrift clients and servers on Netty.
 * [OkHttp](http://square.github.io/okhttp/) - HTTP+SPDY client.
 * [Undertow](http://undertow.io/) - Web server providing both blocking and non-blocking API’s based on NIO. Used as a network layer in WildFly.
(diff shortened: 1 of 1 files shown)


akullpp:
Relatively new library and you are the author right? So why should we add it? Is it the only library that does parsing and serialization according to RFC 2141?

claussni:
That is correct, I'm the author and just published the library.

There is one library I'm aware of, that handles URNs: https://github.com/jcabi/jcabi-urn. However, they claim that the library is not fully compliant to RFC 2141 and might become so in future versions. But it doesn't look like this is going to happen any time soon. I opened a PR in 2014 which is still pending...

According to awesome-java's contribution guidelines, one could say 'urnlib' is a "a niche product that fills a gap". I explicitly wanted it to be small, focused on implementing RFC 2141 only and having no dependencies.

PR #343, opened 2016-04-07
add missing chapter Performance Analysis to the header

EgeniyKiselev:
Performance Analysis chapter was missing in the header list


akullpp:
Thanks, saw your PR too late and added it myself due to the Flight Recorder PR.

PR #347, opened 2016-04-19
Added two distinct library in Imagery section.

mkghosh:


akullpp:
On hold: #351

akullpp:
See #351

PR #353, opened 2016-05-08
move spring to "Dependency Injection" section

### LINKED ISSUE #346: Spring framework category
Hi,

Thank you for this really awesome list!

Just a note about Spring framework (the core). Currently, it is placed under "Web Frameworks" category, I do believe it fits better under "Dependency Injection".

It is more precise to put "Spring MVC" under "Web frameworks" and move the Spring framework (link to core) to "Dependency Injection".

What do you think? I can send a pull request if you agree.

Kind regards
Mahmoud

benas:
this update fixes #346

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [Dagger2](http://google.github.io/dagger/) - Compile-time injection framework without reflection.
 * [Guice](https://github.com/google/guice) - Lightweight but powerful framework that completes Dagger.
 * [HK2](https://hk2.java.net) - Light-weight and dynamic dependency injection framework.
+* [Spring](http://projects.spring.io/spring-framework/) - Provides many packages ranging from dependency injection to aspect-oriented programming to security.
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


akullpp:
But is it correct to place it in Dependency Injection? DI is only one part of the Spring framework, although the original and fundamental one.

sta-szek:
I think Spring framework fits DI as much as Messaging, Security, etc...
Probably there is a need for "Core" section or even "JEE development"

krisskross:
I think having just "Spring" in the DI section make sense. Other Spring modules can be elaborated as such like "Spring Boot" etc, etc.

benas:
I join @sta-szek and @krisskross , the core of the spring framework fits better under DI.

In regards to security, since there is a separate [spring project](http://projects.spring.io/spring-security/) for that, it would be better to place it under [Security](https://github.com/akullpp/awesome-java#security) section. What do you think?

sta-szek:
According to [Spring Framework Project](http://projects.spring.io/spring-framework/) it supports

> Core support for dependency injection, transaction management, web applications, data access, messaging, testing and more.

and

> Features
Dependency Injection
Aspect-Oriented Programming including Spring's declarative transaction management
Spring MVC web application and RESTful web service framework
Foundational support for JDBC, JPA, JMS
Much more…

So placing it in Security or DI section does not make anyone happy.
If you mention about Spring, I think there should be separated section "Spring" and there should be all [Spring Projects](https://spring.io/projects).

benas:
* Spring **core** is the defacto DI container. Even though it provides AOP and other features (and since there is no "AOP" section in the list), the better place for it is under DI section IMO.

* Spring MVC should be placed under "Web Frameworks"

* Spring Security can be used outside spring like an alternative to apache shiro. So IMO it is better to add it under "Security" section

sgwerder:
I'm against splitting up Spring. However then there'd have to be a new category like "Platform" or give Spring it's own category like @sta-szek said.

akullpp:
A general category like `Platform` could be useful since we had platform projects in the past that would be really inconvenient to be split up into separate categories.

On the other hand the granularity of adding popular parts like Spring Boot is really nice, in this case we would need to add it to the description.

The idea of creating a separate `Spring` category could be difficult since it could explode if other platforms arise.

I'm on the fence: either we add a `Platform` category or split it up.

sta-szek:
The platform makes sense to me

akullpp:
To platform it is then.

PR #357, opened 2016-05-27
Orbit URL has changed

JoeHegarty:
Hi, 
I'm the maintainer of the orbit framework and we recently changed primary URLs so submitting this update.

I was also wondering if it's possible to update the source code link at https://java.libhunt.com/project/orbit. 
I realise you don't own this so tagging @StanBright to see how that works. 
It should be https://github.com/orbit/orbit. So the site accurately reflects activity etc.

Thanks!
Joe


StanBright:
@JoeHegarty, thanks for the update. I've updated Orbit @ LibHunt now.

PR #360, opened 2016-06-06
Update README.md

DaveJarvis:
Reference to jARVEST.


akullpp:
Latest news are from 2013, why should we add this library if it's not in active development?

DaveJarvis:
There are lots of libraries out there that are in a stable state that don't need continuous development, which also happen to be awesome. If the requirement for addition to the list is that it is under active development, then (1) state that somewhere; and (2) ignore the pull request.

PR #362, opened 2016-06-07
Update README.md

DaveJarvis:
Added reference to Web Harvest data mining tool.


akullpp:
Latest news and probably release was from 2010, why should we add this library?

DaveJarvis:
There are lots of libraries out there that are in a stable state that don't need continuous development, which also happen to be awesome. If the requirement for addition to the list is that it is under active development, then (1) state that somewhere; and (2) ignore the pull request.

PR #382, opened 2016-07-13
Create a new Caching section, and add Caffeine & move Ehcache to it.

jbduncan:
According to their [GitHub page](https://github.com/ben-manes/caffeine), Caffeine is a highly performant caching library for Java 8 which takes lessons from [Guava's Cache](https://github.com/google/guava/wiki/CachesExplained) and [ConcurrentLinkedHashMap](https://code.google.com/p/concurrentlinkedhashmap). It's so fine-tuned performance-wise that there are early discussions going on to replace caches in [HBase](https://issues.apache.org/jira/browse/HBASE-15560), [Cassandra](https://issues.apache.org/jira/browse/CASSANDRA-10855), [Solr](https://issues.apache.org/jira/browse/SOLR-8241) and [ElasticSearch](https://github.com/elastic/elasticsearch/issues/16802) with Caffeine. Therefore I believe it's worth adding to the list.

Since none of the existing sections seem to match well with Caffeine, I decided to add a new Caching section and move Ehcache to this new section as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ A curated list of awesome Java frameworks, libraries and software.
     - [Bean Mapping](#bean-mapping)
     - [Build](#build)
     - [Bytecode Manipulation](#bytecode-manipulation)
+    - [Caching](#caching)
     - [Cluster Management](#cluster-management)
     - [Code Analysis](#code-analysis)
     - [Code Coverage](#code-coverage)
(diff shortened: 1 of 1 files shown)


ben-manes:
Thanks!

FYI, Ehcache 3 is an MRU which results in a very poor hit rate. I guess that encourages sales by coercing users into adopting the distributed and increasing license cost by requiring more servers. I may be biased, but not a fan of such manipulative practices.

jbduncan:
You're very welcome @ben-manes. :smile:

Out of curiosity, do you have a link that shows that Ehcache's distributed side requires a paid-for license? Admittedly, I've only just skimmed the website,, but I couldn't find any info that verifies what you've said.

ben-manes:
I'm pretty sure I read somewhere the Terracotta server a was free to use as a single instance (no striping, replication, etc). The 2.x series had support for JGroups, etc. from prior to the acquisition. But you're right, I can't find that now so I might have been wrong and misattributed their hostility towards fixing performance problems.

ben-manes:
Thanks for challenging me on that! It really annoyed me thinking they're reason for not fixing these issues (when repeatedly brought to their attention) was malicious, given various other details. But now that I can't find the pricing info it doesn't frustrate me, as could just be personality conflicts. It actually really gnawed to see their talks knowing the data disproved claims, but now I can find other excuses and not feel bothered by it.

PR #396, opened 2016-07-29
Add bluetooth tool

cltsp:


yegeniy:
Android has some bluetooth APIs. Not sure how relevant that is for here, but I assume that's a common use case. 

https://developer.android.com/guide/topics/connectivity/bluetooth-le.html

akullpp:
Unless you have a second entry, please don't open a category. Add another entry or move it to the Misc category.

This list doesn't allow Android-specifc libraries.

cltsp:
@akullpp  I'm sorry to clearly described,but is not Android libraries.
 
This a pc in the library, or application raspberry pi.

"BlueCove is a Java SE library for Bluetooth JSR-82 implementation. runing in Windows,Linux and OSX".

PR #408, opened 2016-08-31
Moved Faux Pas from Functional Programming to Development section

jbgi:
This library is not related to functional programming. Having it in this section might lead learners of functional programming to think that error propagation/handling via runtime exceptions is the way to go. It is not. Also, sneaky throw is a hack/corruption of the type system.
A key goal of functional programming is maximize safety, this library has the opposite goal by encouraging unsafe functions.

In functional programming one would rather use sum types (or at least checked exceptions) for error handling.


akullpp:
Could you please move it to a more suitable category or justify the complete deletion?

PR #417, opened 2016-09-09
Add link for the functional library underscore-java.

javadev:


akullpp:
I'm a bit concerned adding this library. I could understand if you use it Pre-Java 8 but I wouldn't use it when you have streams. Any other opinions?

javadev:
This library was designed for Java 6. It works similar for [Guava].

javadev:
Hi,

```
I'm a bit concerned adding this library. I could understand if you use it Pre-Java 8 but I wouldn't use it when you have streams. Any other opinions?
```

Could you rephrase yiur question please?

Kind regards, Valentyn

javadev:
!

yegeniy:
@javadev what akullpp is saying is that the [Java 8 release](https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html) provides [a set of libraries](https://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html) for the functional tools that underscore.js has. Underscore.js is surely going to be more fully featured, but there is a good argument to sticking with the Java standard library as a general approach.

yegeniy:
Actually, this issue in `underscore-java` is a duplicate of the above concern: https://github.com/javadev/underscore-java/issues/72

yegeniy:
By the way, that's an awesome alias @javadev

javadev:
Short examples for underscore-java methods:

```java
$.first(asList(5, 4, 3, 2, 1), 2);
=> [5, 4]

$.initial(asList(5, 4, 3, 2, 1));
=> [5, 4, 3, 2]

$.last(asList(5, 4, 3, 2, 1), 2);
=> [2, 1]

$.rest(asList(5, 4, 3, 2, 1));
=> [4, 3, 2, 1]
```

@yegeniy How to implement them with help of java 8 stream API?

Kind regards, Valentyn

jbduncan:
@javadev, for the examples you've given, the most practical way to implement them is using `List.subList`, as you've already done in the source code of underscore-java.

It's possible to use `Stream` to replace `$.first` and `$.rest` (as shown below), but then their behaviour changes a bit.
```java
asList(5, 4, 3, 2, 1).stream().limit(2).collect(Collectors.toList());
=> [5, 4]
```
```java
asList(5, 4, 3, 2, 1).stream().skip(1).collect(Collectors.toList());
=> [4, 3, 2, 1]
```

These examples differ from `$.first` and `$.rest` in that they produce _entirely new lists_ rather than _views_ of the original list. This may or may not be what you want. 

The 1st example with `limit` is relatively efficient at producing a new list, because the `limit` method itself is _short-circuiting_, which means it effectively moves onto the `collect` method as soon as 2 elements in the stream (not all of them) have been traversed.

The 2nd example is not so efficient, because `skip` does _not_ short-circuit; that is, it _has_ to traverse over all the elements in the stream to return the elements after the first one.

But despite these performance and behaviour differences, `Stream` does have a large advantage - you can produce a stream over an `Iterable` or some other "collection" that doesn't have a fixed size. This is useful if you don't want to copy the iterable into a list but want to perform functional operations on it, e.g.
```java
Iterable<Integer> ages = getAgesFromDatabaseLazily();
StreamSupport.stream(ages.spliterator(), false).limit(2).mapToInt(age -> age + 1).forEach(System.out::println);
```

As for `$.initial` and `$.last`, as far as I can tell, those methods **must** operate on indexed collections like `List`, because as you're traversing over a `Stream` there's no way to tell in advance how many elements are left (just like with an `Iterable`).

**TL;DR** `Stream` isn't a complete replacement for underscore-java, but they can both complement each other.

javadev:
!

jbduncan:
Sorry, was that too much information for you? :)

javadev:
https://github.com/Vedenin/useful-java-links added undersore-java. Wow.

screen shot 2016-09-14 at 2 55 11 pm

README.md
```diff
@@ -748,6 +748,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [JADE](http://jade.tilab.com/) - Framework and environment for building and to debugging multi-agent systems.
 * [JavaVerbalExpressions](https://github.com/VerbalExpressions/JavaVerbalExpressions) - A library that helps to construct difficult regular expressions.
 * [Protégé](http://protege.stanford.edu/) - Provides an ontology editor and a framework to build knowledge-based systems.
+* [Underscore-java](http://javadev.github.io/underscore-java/) - A java port of Underscore.js. In addition to porting Underscore's functionality, Underscore-java includes matching unit tests.
```
akullpp: Please fix the description:

`Port of Underscore.js functions.`

No need to mention tests, they are a assumed. The "Java"-part is implicit given this repo.

javadev: 100% code coverage is a key feature of this library. I copied this text from github readme page.

javadev: Changed description to the Port of Underscore.js functions.
Thanks.

PR #421, opened 2016-09-27
Update README.md

dreanor:
added cruisecontrol


akullpp:
The description is really lacking and the tool itself hasn't been updated since 2010.

PR #423, opened 2016-10-04
Added OrientDB to distributed database

cdedeurwaerder:


akullpp:
How is it related to Java the programming language other than it has a Java driver? Bear in mind that we don't add tools written in Java and it would be disputable to add every DB which has drivers for Java.

cdedeurwaerder:
OrientDB is written in java and is embeddable in java application !

PR #430, opened 2016-10-15
Etcd

jonatan-ivanov:
Distributed reliable key-value store for the most critical data of a distributed system: https://coreos.com/etcd/


jbduncan:
I think it would make more grammatical sense to say "**A reliable, distributed** key-value store..." instead of "**Distributed reliable** key-value store".

jbduncan:
Oh, and also I just noticed that etcd is written on Go, rather than Java.

Could you explain why you think it's a good fit for awesome-java?

akullpp:
Why "awesome Java"?

jbduncan:
@akullpp was your question directed at me or at @jonatan-ivanov? :)

akullpp:
I don't see an explicit relation to Java

PR #432, opened 2016-10-23
Adding Java Off Heap podcast

tginsberg:
This podcast is run by members of the Chicago Java Users Group and
has new episodes every month. Each episode is around an hour long and
covers current events in the Java community.


akullpp:
Nice, you could add a description though ;-)

tginsberg:
Sorry about that, I'll provide a description in a subsequent PR.

PR #433, opened 2016-10-27
Add Membrane Service Proxy

helpermethod:


akullpp:
I'm going to evaluate the entry at a later point.

helpermethod:
Hi @akullpp,

I've added Membrane Service Proxy because it

* can be embedded into a Java application
* can be ran as a Maven plugin for testing
* can be added as a servlet endpoint of a WAR application
* exposes a REST

PR #438, opened 2016-10-30
add benchmark module and jmh

cyberdak:
add benchmark module and jmh


akullpp:
How does the new section differ from the already existing **Performance Analysis** section?

cyberdak:
Reading Performance Analysis section's description,it inculde benchmarking.
So please close this PR.I will submit a new one which insert JMH to the Performance Analysis section.

PR #441, opened 2016-11-05
Update README.md

zhengjunbase:
add codehelper.generator


akullpp:
I don't think we should start adding IDE Plugins... any opinions?

karollewandowski:
At first I thought it'd be nice to have list of cool plugins for very popular tools, but it's very hard to define the criteria for what is very popular/cool/useful. Such plugins section can explode soon (it's hard to predict) and just duplicate tools (eg. Mockito in Tests section and Mockito plugin in Plugins section). Also all major IDEs have plugins repositories:
- https://plugins.jetbrains.com/
- https://marketplace.eclipse.org/
- http://plugins.netbeans.org/

with search and sections like "most popular". Maybe it'd be good to add these links next to IDE descriptions. Another idea is to mention plugins next to tool, but it should be rather mentioned on tool's site, so I think it's not a good one.
Summary: I think it's bad idea, but link to plugins repository next to IDE would be nice.

PR #444, opened 2016-11-21
Added Facebook Infer

codehate:
Added (Facebook Infer](http://fbinfer.com/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [PMD](https://github.com/pmd/pmd) - Source code analysis for finding bad coding practices.
 * [SonarQube](http://www.sonarqube.org/) - Integrates other analysis components via plugins and provides an overview of the metrics over time.
 * [Spoon](https://github.com/INRIA/spoon/) - Library for analyzing and transforming Java source code.
+* [Infer](http://fbinfer.com/) - Tool to produce a list of potential bugs.
 
 ## Code Coverage
 


jbduncan:
Currently Facebook Infer only targets Android (as far as Java's concerned), so I'd personally hesitate to have it listed in awesome-java.

My reason for being hesitant is because awesome-java used to list [Picasso](http://square.github.io/picasso/) as an awesome library, but eventually it was removed because it could only be used on Android, and awesome-java is supposed to be a place for libraries that can be used on most or all flavours of Java.

codehate:
@jbduncan I doubt your thought that Infer only targets Android, If you have a look at Infer homepage, the [preview gif shown clearly targets Java](https://asciinema.org/api/asciicasts/32101?speed=2&autoplay=true&loop=true&t=23), and [Infer's about page](http://fbinfer.com/docs/about-Infer.html) states:

> Infer is a static program analyzer for **Java**, C, and Objective-C, written in OCaml.

jbduncan:
Oh, my apologies @codehate, it seems I spoke too soon! Thanks for correcting my misunderstanding on Infer's applicability to Java code in general. :+1:

In that case, I think it's worth considering including this in awesome-java.

@akullpp, thoughts?

akullpp:
The page clearly states that the use case is for mobile apps.

jbduncan:
Well, yes, they do state that on the main page, but I think it's a misleading description.

I dug into the [about page](http://fbinfer.com/docs/about-Infer.html) just now, and it explains that although Infer is only used at Facebook for their Android and iOS apps, it can also be used to analyse C code and bog-standard Java programs that can be compiled with `javac`. See [this preview gif on the homepage](https://asciinema.org/api/asciicasts/32101?speed=2&autoplay=true&loop=true&t=23) for an example of how Infer can analyse normal Java code.

So the way I see it, Infer actually works a bit like Google's error-prone tool or FindBugs, the main difference being it uses a couple of novel (i.e. very recent) software verification techniques called **separation logic** and **bi-abduction** to do the static analysis.

codehate:
@jbduncan Great! Thanks for the explanation. So, are you certain to add Infer to awesome-java?

jbduncan:
Hi @codehate, I actually don't have write access to awesome-java, so you'll have to wait patiently for @akullpp to have his say on whether Infer is sufficiently Java-orientated and awesome enough to be added to the list. :)

codehate:
@jbduncan Sure, let's wait for @akullpp call

PR #446, opened 2016-11-22
add gceasy.io

cyberdak:
add gceasy.io which is a web gc log analyser.


asgs:
This is a very good tool! thanks for adding

akullpp:
I've added it manually since there was a merge conflict.

Issue #384, opened 2016-07-16
why not zbus


WenTao-Love:
ZBUS=MQ+RPC, an extremely light-weighted(~300K) Service Bus focus on Message Queue, Remote Procedure Call, HTTP MicroService, capable of tackling DMZ network constrains, with multiple languages(Java/C#/C/Python/Javascript) support. http://zbus.org 
!!!!!!

jbduncan:
I can't speak on behalf of @akullpp, but I know for me at least, I'd struggle to consider zbus for inclusion in awesome-java, because your link currently leads to a non-English website, meaning I can't understand what zbus is about without using Google Translate (which isn't 100% reliable even at the best of times).

Does zbus have an proper English website we can refer to?

jbduncan:
Not to mention, even when I use Google Translate myself, the website is... barebones, to say the least.

akullpp:
As @jbduncan mentioned I don't add sites that cater only to specific language speakers (Chinese the most common example). English is the lingua franca.

PR #457, opened 2016-12-11
Added hprose

andot:


pacher:
I want to point out that the docs(wiki) are only in Chinese\Japanese\Korean, no javadocs, so it does not seem to be useful for the rest of the world. May be some local\localized awesome list is the better place for it?

P.S. www.hprose.com and www.hprose.org don't have english version

andot:
It's an RPC library, the javadocs can't describe how to use it, so we're going to provide a separate document and tutorial to introduce how to use it, but now, it only has Chinese documents, there's no one to help us to translate it to English, but later maybe.

akullpp:
Please refer to https://github.com/akullpp/awesome-java/issues/384#issuecomment-233122944

If you provide a translation in English, I'd be very happy to add it.

andot:
@akullpp, Thank you, I will add English document and tutorial later.

PR #460, opened 2017-01-06
Added a new java framework

m0ver:


bwajtr:
Hm, I'm against merging this into master for these reasons:

1. Spring is already in the list (but maybe it's just badly merged master)
2. http://tinystruct.org site is empty now. I have troubles google it or find out what this framework/library does or is good for...
3. I don't think that reference to any repository with examples (of any framework) belongs to this list...

akullpp:
Just as @bwajtr said.

PR #478, opened 2017-02-22
Add Embulk

muga:
Hi, would you consider adding [Embulk](http://www.embulk.org/) ?

Embulk is an open-source bulk data loader that helps data transfer between various databases, storages, file formats, and cloud services. 

Thank you.


akullpp:
Is there an exposed API for Java integration? I ask because this is a list about the programming language Java not about applications written in Java.

muga:
Yes, Embulk provides [plugin API](http://www.embulk.org/docs/javadoc/index.html). It enables loading data using combination of input and output plugins written by the API.

PR #481, opened 2017-02-24
Move Hamcrest to Ancients

heldev:
Last release - 2 years ago. Last commit - 9 month ago.
Now we have other convenient and active libs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [FreeMarker](http://freemarker.org/) - General templating engine without any heavyweight or opinionated dependencies.
 * [GlassFish](https://glassfish.java.net/) - Application server and reference implementation for Java EE sponsored by Oracle.
 * [GWT](http://www.gwtproject.org/) - Toolbox which includes a Java-to-JavaScript compiler for client-side code, XML parser, API for RPC, JUnit integration, internationalization support and widgets for the GUI.
+* [Hamcrest](http://hamcrest.org/JavaHamcrest/) - Matchers that can be combined to create flexible expressions of intent.
 * [HornetQ](http://hornetq.jboss.org/) - Clear, concise, modular and made to be embedded.
 * [Hudson](http://hudson-ci.org/) - Continuous integration server still in active development.
 * [Java Modeling Language (JML)](http://www.eecs.ucf.edu/~leavens/JML/) - Behavioral interface specification language that can be used to specify the behavior of code modules. It combines the design by contract approach of Eiffel and the model-based specification approach of the Larch family of interface specification languages, with some elements of the refinement calculus. Used by several other verification tools.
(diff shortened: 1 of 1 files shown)


bwajtr:
> Now we have other convenient and active libs.

What would you recommend instead of Hamcrest for "readable assertions"? Note it's ok for me if it's Java8 only... I found following candidates:
[Truth](https://github.com/google/truth)
[AssertJ](http://joel-costigliola.github.io/assertj/)

Do you know of any others?

heldev:
I would recommend AssertJ, as I know it's the most advanced library now. But I have to admit, I didn't try Truth. When I was looking for alternatives, Truth was quite "raw" (0.* version, not much info about it, no Joda Time support). And as I can see, nothing has changed yet. 


> other convenient and active libs
Even the ancestor of AssertJ [FEST](https://github.com/alexruiz/fest-assert-2.x) is more readable and convenient than Hamcrest.

jbduncan:
I agree that AssertJ is currently more advanced, but by my understanding, Truth's main advantage in comparison to AssertJ is that it's highly extensible - one can create something called a `Subject` which allows them to assert on types which aren't covered by Truth main, like Joda Time's date classes, and to then write stuff like
```java
assertAbout(JODA_DATA_TIME_SUBJECT).that(new DateTime().withYear(2017)).hasYear(2017);
```

It also seems that Truth's documentation has also improved a lot recently; see http://google.github.io/truth/faq and http://google.github.io/truth/extension for more info if you're interested @heldev. :)

(They've also written up a sort of comparison list between the JUnit, AssertJ and Truth APIs for simple assertion examples at http://google.github.io/truth/comparison).

jbduncan:
@bwajtr, I can't speak for @heldev, but I'm not personally aware of any other alternatives to AssertJ and Truth, and so I'd be inclined to list them both as good replacements for Hamcrest. :)

heldev:
@jbduncan, 

> main advantage in comparison to AssertJ is that it's highly extensible

I took a look and didn't find the advantage, It looks they are equally extensible. AssertJ even has automated [tools](http://joel-costigliola.github.io/assertj/assertj-assertions-generator.html) to generate class specific assertions.

jbduncan:
> AssertJ even has automated tools to generate class specific assertions.

Oh yes, you're right! Interesting! Thanks for correcting my misunderstanding here. :)

In that case, AFAICT, there are only two other areas where Truth might be more useful than AssertJ:
1. One can write all assertions in the fluent style with Truth, whereas it's not entirely possible to do so with AssertJ. For example, in AssertJ, one has to write something like the following to assert equality for doubles:
```java
assertThat(actualDouble).isCloseTo(expectedDouble, offset(offset));
```
Whereas in Truth one can write:
```java
assertThat(actualDouble).isWithin(tolerance).of(expectedDouble);
```
2. Truth can do both JUnit-style assumptions (skipping behaviour) and expectations (fail-at-end behaviour), whereas AssertJ can only do expectations. E.g. in AssertJ one can only write:
```java
softly.assertThat(expected).isEqualTo(actual); // @Rule for expectations with JUnitSoftAssertions
```

Whereas in Truth one can write:
```java
assume().that(expected).isEqualTo(actual);
expect.that(expected).isEqualTo(actual); // Supplied by JUnit @Rule
```

heldev:
Yes, it's true - AssertJ can't do it OOTB

bwajtr:
Thanks all...
Just found out that AssertJ is included in the spring boot testing starter... so it's a quite simple choice for me now :)

PR #495, opened 2017-03-29
added Governator

liviutudor:


jbduncan:
Nebula and the Gradle Docker plugin are more appropriate for [awesome-gradle](https://github.com/ksoichiro/awesome-gradle) I think. :)

liviutudor:
Oh I was so not aware of that! Thank you I will add them there and remove them from here.
Is Governator though still a valid entry? Shall I push an update with just that?
Thanks for taking the time to review this btw!

liviutudor:
I've updated this just in case you still think it's worthy

jbduncan:
I think the description for Governator is too long. How about something like this?
> Extensions and utilities that enhance Google Guice.

What do you think @akullpp?

liviutudor:
agreed, made the change.

PR #496, opened 2017-04-04
Remove Spock because it's a Groovy framework

heldev:
Spock is a nice tool, but it has the same rights to be here as, for example, ScalaCheck.

### DIFF
--- a/README.md
+++ b/README.md
@@ -787,7 +787,6 @@ A curated list of awesome Java frameworks, libraries and software.
 * [REST Assured](https://github.com/jayway/rest-assured) - Java DSL for easy testing for REST/HTTP services.
 * [Selenide](http://selenide.org/) - Concise API around Selenium to write stable and readable UI tests.
 * [Selenium](http://docs.seleniumhq.org/) - Portable software testing framework for web applications.
-* [Spock](http://spockframework.github.io/spock/docs/) - JUnit-compatible framework featuring an expressive Groovy-derived specification language.
 * [TestContainers](https://github.com/testcontainers/testcontainers-java) - Provides throwaway instances of common databases, Selenium web browsers, or anything else that can run in a Docker container.
 * [Truth](https://github.com/google/truth) - Google's assertion and proposition framework.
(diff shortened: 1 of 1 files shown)


akullpp:
Interesting proposal. I used it before and I was quite happy with it, so I'm a bit biased. But I tend to agree with you.

It uses Groovy as a language for testing which is not uncommon, e.g. some people use CoffeeScript for frontend testing since it is less verbose. The same applies here.

Could your argument be valid for other cases too? Should we delete Gradle for example?

README.md
```diff
@@ -542,6 +542,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [Modern Java - A Guide to Java 8](https://github.com/winterbe/java8-tutorial) - Popular Java 8 guide.
 * [Modernizer](https://github.com/andrewgaul/modernizer-maven-plugin) - Detect uses of legacy Java APIs.
 * [OpenRefine](http://openrefine.org/) - Tool for working with messy data: cleaning, transforming, extending it with web services and linking it to databases.
+* [Polyglot for Maven](https://github.com/takari/polyglot-maven/) - Set of extensions for Maven 3.3.1+ that allows the POM model to be written in dialects other than XML.
```
akullpp: Please create another PR and don't do Add/Delete in a single one since deletes are much more likely to be open for discussion.

README.md
```diff
@@ -786,7 +787,6 @@ A curated list of awesome Java frameworks, libraries and software.
 * [REST Assured](https://github.com/jayway/rest-assured) - Java DSL for easy testing for REST/HTTP services.
 * [Selenide](http://selenide.org/) - Concise API around Selenium to write stable and readable UI tests.
 * [Selenium](http://docs.seleniumhq.org/) - Portable software testing framework for web applications.
-* [Spock](http://spockframework.github.io/spock/docs/) - JUnit-compatible framework featuring an expressive Groovy-derived specification language.
```
akullpp: We have a strange description anyway. JUnit-compatible is wrong, I think it uses the JUnit runner though.

PR #499, opened 2017-04-13
Add schemaspy - java based database documentation tool

wkasa:
SchemaSpy is a Java-based tool (requires Java 8 or higher) that analyzes the metadata of a schema in a database and generates a visual representation of it in a browser-displayable format. It lets you click through the hierarchy of database tables via child and parent table relationships as represented by both HTML links and entity-relationship diagrams. It’s also designed to help resolve the obtuse errors that a database sometimes gives related to failures due to constraints. It is free software that is distributed under the terms of the MIT License.
http://schemaspy.org/


akullpp:
How is it related to Java?

wkasa:
@akullpp - it is written in java so according to description: "A curated list of awesome Java frameworks, libraries and software" I though that you may consider to add this software to your list.

akullpp:
The logic doesn't apply else we would have to add every software written in Java. Tools without API are generally only added if they have a specific Java-related scope, like Performance Analysis tools.

PR #500, opened 2017-04-16
rename Javaslang to Vavr

agebhar1:
Due to [1] »Javaslang« was renamed to »Vavr« on 14th April 2017.

[1] http://blog.vavr.io/javaslang-changes-name-to-vavr/


rbjorklin:
Maybe add "Formerly Javaslang" ?

PR #506, opened 2017-04-24
Add ArchUnit

koppor:
[ArchUnit](https://github.com/TNG/ArchUnit) is unique in its approach. It provides a method to add unit tests on the architecture of a system based on a fluent API.

The alternative is [Macker](https://innig.net/macker/index.html), which also implements architecture checks. Example: https://innig.net/macker/example/layering/src/macker.xml. However, it is very outdated and thus not awesome.


akullpp:
Isn't it a testing library and should it be in the **Testing** section?

koppor:
Sure. Fixed.

PR #510, opened 2017-04-30
add cdp4j

ui4j:


jbduncan:
I would struggle to call this an awesome Java project, for a couple of reasons:
1. It's only been open sourced for 9 hours at the time of me writing this, so it's not had time to prove itself in the wider Java community.
2. I'd argue it's not even a true open source project, since one can only submit issues & bug reports **if they pay for a subscription**. AFAIK this has never been done before, and on a personal level it makes me feel very iffy.

akullpp:
@jbduncan I agree, let's see how this works out. A small pitch based on the unique selling points would be nice, too.

akullpp:
Closed due to inactivity.

PR #513, opened 2017-05-02
added nebula and freeline

Moshe-Immerman:


jbduncan:
And I'm not so sure that Freeline is a good fit for awesome-java either, as it seems to be very Android-specific.

PR #528, opened 2017-05-12
Submitting Mines Java Toolkit under Science

chrisengelsma:


akullpp:
See contribution rules:

There are two required criteria for a pull request:

1. If an entry has a similar scope as other entries in the same category, the description must state the unique features that distinguishes it from the other entries.

2. If an entry does not meet conditions (a) to (d) there has to be an explanation either in the description or the pull request why it should be added to the list.

chrisengelsma:
Updated to reflect unique attributes of this library

akullpp:
Let's add it since its really niche. We will see how it holds up and if it goes the way of all scientific libraries for Java...

PR #529, opened 2017-05-13
Added related awesome resource section

Moshe-Immerman:


jbduncan:
I think this is a good idea. An awesome-maven list could be added as well (if it exists), so that we can move existing entries like "really-executable-jars-maven-plugin" and "Maven for Polyglot" to a more appropriate place and still have a link to it. :)

PR #531, opened 2017-05-16
Remove baratine

Moshe-Immerman:
GPL license is not awesome for libraries, Only 2 contributors, and 50 stars


jbduncan:
I agree with this removal.

What do you think, @akullpp?

PR #532, opened 2017-05-16
Clarify contribution guidelines

Moshe-Immerman:


jbduncan:
I think it should say the following.

> Unfortunately, not every library/tool/framework can be considered. The aim of the list is to provide a concise list of noteworthy modern software. This means that suggested software meets at least one of the following criteria:
>
> (a) ...
> ...
> (d) ...
>
> It should also be developer-friendly (for example, it should have a business-friendly open source license or, if commercial in nature, have clear pricing). However, this is not strictly necessary as long as it meets one of the criteria above; we do occasionally accept GPL-licensed projects, but we will consider them more critically.

What do you think, @akullpp?

jbduncan:
@Moshe-Immerman "Github" -> "GitHub"

Moshe-Immerman:
my bad, I missed the emphasis on library vs application.

jbduncan:
@Moshe-Immerman Looks much better, many thanks. :)

jbduncan:
Waiting on @akullpp's opinion now.

PR #533, opened 2017-05-16
move Akka to reactive

Moshe-Immerman:
While it can (and is used for distributed apps) it's defining features are reactive
[http://doc.akka.io/docs/akka/2.5.1/intro/what-is-akka.html](http://doc.akka.io/docs/akka/2.5.1/intro/what-is-akka.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,7 +309,6 @@ A curated list of awesome Java frameworks, libraries and software.
 
 *Libraries and frameworks for writing distributed and fault-tolerant applications.*
 
-* [Akka](http://akka.io) - Toolkit and runtime for building concurrent, distributed, and fault tolerant event-driven applications.
 * [Apache Storm](http://storm.apache.org/) - Realtime computation system.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -676,6 +675,7 @@ A curated list of awesome Java frameworks, libraries and software.
 
 *Libraries for developing reactive applications.*
 
+* [Akka](http://akka.io) - Toolkit and runtime for building concurrent, distributed, and fault tolerant event-driven applications.
```
jbduncan: The description should probably say something like, "Toolkit and runtime for building concurrent, reactive, distributed, and fault tolerant event-driven applications."

jbduncan:
I'm in two minds about this. On one hand, it does indeed say in http://doc.akka.io/docs/akka/2.5.1/intro/what-is-akka.html that it's "reactive", but it states on the main page that it "is a toolkit and runtime for building highly concurrent, distributed, and resilient message-driven applications on the JVM". Consequently I can't decide which category it should go under.

akullpp:
Very difficult. I think the category itself is flawed and needs to be reworked.

PR #534, opened 2017-05-16
Move multiple distributed libraries

Moshe-Immerman:


README.md
```diff
@@ -309,18 +309,24 @@ A curated list of awesome Java frameworks, libraries and software.
 
 *Libraries and frameworks for writing distributed and fault-tolerant applications.*
 
+
```
jbduncan: I think this extra line is simply not needed. :)

akullpp:
This PR is really problematic, many of the libraries state explicitly that they are used for a microservice architecture and are probably not as viable with a monolithic approach.

PR #535, opened 2017-05-16
Remove SnopEE

Moshe-Immerman:
I think this is a bit immature to be awesome, the library is self-confessed experimental The development also seems to have split into 2 repos: https://github.com/ivargrimstad/snoop and https://github.com/ivargrimstad/snoop**ee**


jbduncan:
Sounds like a very valid reason to have this removed.

What do you think @akullpp?

PR #536, opened 2017-05-16
Remove REST languages

Moshe-Immerman:
These aren't java and are better suited to an awesome REST list - #529 adds a link to one


jbduncan:
I think this is a good idea, combined with #529.

What do you think @akullpp?

PR #543, opened 2017-05-25
Add awesome microservices link

Moshe-Immerman:
I think some of the non-java specific tools/libraries can be removed after this

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ A curated list of awesome Java frameworks, libraries and software.
     - [Build](#build)
     - [Bytecode Manipulation](#bytecode-manipulation)
     - [Caching](#caching)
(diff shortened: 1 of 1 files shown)


akullpp:
Could you pease remove the non java-specific tools mentioned?

akullpp:
You removed the entire cluster management section but the "Awesome Microservices" list doesn't include them. I don't know if they would accept an PR adding the removed tools since they are not necessarily microservice focused. So how would someone no about these awesome tools if they are not mentioned anywhere?

Moshe-Immerman:
@akullpp I have created a new PR just to add the link to awesome microservices, and I will wait for https://github.com/mfornos/awesome-microservices/pull/93 to either merge or close before re looking at removing items

PR #545, opened 2017-05-25
Remove distributed databases

Moshe-Immerman:
awesome microservices has a similar list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ A curated list of awesome Java frameworks, libraries and software.
     - [Dependency Injection](#dependency-injection)
     - [Development](#development)
     - [Distributed Applications](#distributed-applications)
-    - [Distributed Databases](#distributed-databases)
     - [Distribution](#distribution)
     - [Document Processing](#document-processing)
     - [Formal Verification](#formal-verification)
@@ -134,6 +133,7 @@ A curated list of awesome Java frameworks, libraries and software.
 
 * [Caffeine](https://github.com/ben-manes/caffeine) - High-performance, near-optimal caching library.
 * [Ehcache](http://www.ehcache.org/) - Distributed general-purpose cache.
+* [Infinispan](http://infinispan.org/) - Highly concurrent key/value datastore used for caching.
 
 ## Cluster Management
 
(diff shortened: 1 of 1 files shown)


jbduncan:
I disagree with this change. AFAICT, most or all of the entries under Distributed Databases are implemented in Java.

jbduncan:
Hmm, I think I might've spoke too soon. Considering https://github.com/akullpp/awesome-java/pull/543, I'm now on in two minds about this change. :thinking:

jbduncan:
My gut feeling now is that anything which is at least half implemented in Java deserves to stay.

Moshe-Immerman:
I think this list is targeted at Java developers - So unless you can **use** it from Java it shouldn't be here. 

Cassandra, druid and OpenTSB don't even mention java on the homepage. 

HBase is perhaps marginal as only has official Java API, but there are plenty clients in other languages. 

Infinispan **can** be run embedded, but it kind of defeats the purpose of a distributed cache

Moshe-Immerman:
If anything Infinispan should be moving to the caching category

jbduncan:
@Moshe-Immerman That's a good point! I'm more convinced now that the distributed databases aren't a good fit for awesome-java.

Yeah, your point about Infinispan potentially moving to Caching makes sense, considering that it is available as an embedded Java library (even if it kind of defeats the point of it being distributed) and the fact that it [implements the JCache standard](http://infinispan.org/tutorials/simple/jcache/).

jbduncan:
On the topic of HBase, I can't tell if it deserves to stay, as I don't know how often Java developers use it outside the Hadoop ecosystem.

And talking about Hadoop, I think that Hadoop itself _should_ stay (it's currently in the Ancients section), because IIRC one can write Hadoop MapReduce jobs in Java, even if it's not the most popular language for doing so AFAIK.

jbduncan:
LGTM @Moshe-Immerman!

Before I merge this, would you mind squashing the two commits together, so that when I do merge it, the Git history looks a bit cleaner. :)

Moshe-Immerman:
@jbduncan  done

jbduncan:
@Moshe-Immerman Thanks!

@akullpp, does this look okay to merge to you?

akullpp:
OpenTSDB has a Java API I think.

Overall it looks good.

PR #546, opened 2017-05-27
Add Facebook infer

Moshe-Immerman:


akullpp:
Was deleted since it isn't really Java-focused, you disagree?

Moshe-Immerman:
PMD isn't entirely java focused anymore either?

I don't think the focus of the project should be the criteria, but rather the awesomeness to Java developers. I don't think any of the other analysis tools can do thread safety checks?

akullpp:
I think the argument is interesting. Couldn't you say something similar about Cassandra for example ("awesomeness to Java developers") which was just removed?

So where do we draw the line?

Moshe-Immerman:
@akullpp I think anything that requires or recommends an installation or deployment step onto a server should be excluded (with the exception of maybe monitoring and distribution)

e.g. Cassandra, Hadoop, Zookeeper, Storm, Graylog, Logstash, Kibana, Spark, Kafka should all be removed using that definition.

jbduncan:
Although I am in favour of including Facebook Infer because of its usefulness for Java developers, I've recently re-realised that the punchline at the top of the awesome-java list says,

> A curated list of awesome Java frameworks, libraries and software.

This seems to imply that Infer doesn't fit, because it is written in OCaml.

Perhaps the punchline should be updated?

jbduncan:
P.S. I've realised that "punchline" is the wrong word; please read as "summary". :)

PR #554, opened 2017-06-03
Added Hadi Hariri to Twitter category

LappleApple:
Added Hadi.


akullpp:
He seems rather Kotlin-focused, any thoughts?

PR #562, opened 2017-06-04
Add Zalenium

elgalu:
Zalenium is unique and well supported so worth adding IMHO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,6 +803,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [TestContainers](https://github.com/testcontainers/testcontainers-java) - Provides throwaway instances of common databases, Selenium web browsers, or anything else that can run in a Docker container.
 * [Truth](https://github.com/google/truth) - Google's assertion and proposition framework.
 * [WireMock](http://wiremock.org/) - Stubs and mocks web services.
+* [Zalenium](https://github.com/zalando/zalenium) - Allows anyone to have a disposable and flexible docker-based Selenium Grid infrastructure featuring video recording, live preview and online/offline dashboards.
 
 ## Utility
 


akullpp:
How is this specifically related to the Java? I can see that it's written in Java but isn't it language-agnostic as long as I use Selenium?

Any thoughts on whether we should add general libraries like this or keep the list more focused? @jbduncan

jbduncan:
@akullpp I'm very torn on this. I'm kinda tempted to see how far the idea of "keeping the list more focused" takes us, and if we realise that it leads to a situation where many entries end up removed and/or many PRs end up rejected, then we can loosen a bit and start accepting more general libraries.

We might want to reject this PR and remove Selenium for this reason? But I'm a torn on doing so because on one hand it is useful for Java developers, but on the other it's only partially written in Java (according the its GitHub page) and I think it's useful in general for web developers.

tl;dr, I'm very on the edge about this. :/

jbduncan:
I've just done a search, and it turns out there's an [awesome-selenium](https://github.com/christian-bromann/awesome-selenium) list which contains an entry for Zalenium. So I think we should not accept this PR, but instead, we need to at the very least add a link at the bottom of awesome-java pointing to awesome-selenium, so that we indirectly link to Zalenium.

diemol:
I understand @akullpp's point of view, since the use of Zalenium is not limited exclusively to the Java language (because it is based entirely on Selenium), and after reading his comment I thought that maybe this PR didn't make so much sense.

But then I read @jbduncan's comments, and I got torn about it, since in the Testing section of the awesome list there are other items that can be used from different languages (e.g. Cucumber, Galen, and Selenium itself...). After seeing those cool items on the list, I got the impression that we have awesome things written in Java and they could be language agnostic. 

Anyway, it is just my opinion :), perhaps the solution of linking the other list works.

PS: Thanks for gathering this list and spending time to curate it! I found a few resources that I didn't know about and they might be helpful for us here.

Moshe-Immerman:
I think Galen, Zalenium and Selenium are awesome for all developers and should probably be replaced by the awesome link - Maybe inline in the testing section? 

The link for cucumber is the Java implementation, so I think it should stay.

Just my 2c

jbduncan:
I think @Moshe-Immerman's comment makes a fair amount of sense. Although perhaps, rather than inlining a link to awesome-selenium into **Testing**, a small note like the following could be put instead.

> For more general testing resources, see Awesome Selenium et al. under [Awesome Lists](https://github.com/akullpp/awesome-java#awesome-lists).

jbduncan:
At this point, I (and I'm sure @akullpp as well) would be open to a new PR implementing suggestions from this thread. If anyone is up for that, please feel free to go ahead. :)

akullpp:
I agree, please create a PR. Zalenium is already mentioned in https://github.com/christian-bromann/awesome-selenium

PR #571, opened 2017-06-16
Added NoException and removed Faux Pas

robertvazan:
This change replaces Faux Pas, which is already listed, with NoException. NoException is a superset of Faux Pas functionality with support for all functional interfaces and with complete documentation. NoException is also a superset of functionality found in other similar libraries (listed on [NoException site](https://noexception.machinezoo.com/)). Arguably, NoException also has a cleaner API.

Category was changed from Development to the more appropriate Functional Programming.

Disclaimer: I am the author of NoException.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,6 @@ A curated list of awesome Java frameworks, libraries and software.
 
 * [AspectJ](https://eclipse.org/aspectj/) - Seamless aspect-oriented programming extension.
 * [DCEVM](http://dcevm.github.io/) - Modification of the JVM that allows unlimited redefinition of loaded classes at runtime.
-* [Faux Pas](https://github.com/zalando/faux-pas) - Library that simplifies error handling by circumventing the issue that none of the functional interfaces in the Java Runtime is allowed by default to throw checked exceptions.
 * [HotswapAgent](https://github.com/HotswapProjects/HotswapAgent) - Unlimited runtime class and resource redefinition.
 * [JavaParser](https://github.com/javaparser/javaparser) - Parse, modify and generate Java code.
 * [JavaSymbolSolver](https://github.com/javaparser/javasymbolsolver) - A symbol solver for Java.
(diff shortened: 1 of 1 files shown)


robertvazan:
FYI: NoException was added to [tedyoung/awesome-java8](https://github.com/tedyoung/awesome-java8). AFAIK the two lists are not exclusive, so inclusion here still makes sense.

jbduncan:
I think that both Faux Pas and NoException are poor fits for awesome-java. Allow me to try to explain why.

I'm uncomfortable with how Faux Pas and NoException encourage their users to throw checked exceptions as if they were runtime exceptions (internally using a trick to bypass the compiler's checks). That may be fine if everyone on a greenfield project knows that such trickery is going on and there are top-level handlers to catch and deal with such exceptions (as projects like JUnit 5 do). But if used in a legacy project with many developers, I think it can lead to problems where developers get confused that they're unable to catch checked exceptions because the compiler thinks it's a runtime exception. (This problem is explained in [Lombok's @SneakyThrows](https://projectlombok.org/features/SneakyThrows) doc.)

Am I being too paranoid here?

robertvazan:
@jbduncan I knew before developing NoException that such objections would be raised. That's why NoException has `Exceptions.wrap()` besides `Exceptions.sneak()`. It's your decision, not framework's decision. I wouldn't bash sneaky throws too much though. They make java cool again for folks who would otherwise prefer node.js or ruby. These people then contribute libraries that even enterprises are happy to use.

But more importantly, the primary purpose of NoException is not to handle checked exceptions as those are rare. NoException is primarily intended to be used as a catch-all handler when invoking callbacks/events/jobs or when programming resilient code with fallbacks. NoException returns empty Optional upon exception, which allows for beautiful, functional fallback code. It is a unique feature that I haven't found in any other library. I haven't even seen it discussed anywhere.

NoException encourages you and provides tools for you to define your own reusable exception handlers to fit your project's policies. NoException therefore actually supports consistent exception handling within projects. Even if you have rogue developers on the team, you will find it much easier to search the code for `Exceptions.sneak()` than searching for `try { ... } catch { /* some no-op code * / }` and many of its variants.

robertvazan:
@jbduncan I just checked and lombok is listed too. Perhaps this issue was discussed before.

jbduncan:
Awrigh', you've convinced me. :wink:

If I don't hear back from @akullpp in a couple of days, I'll assume he's fine with this and I'll merge it in. :)

akullpp:
I have 2 issues.

1. This library should not be listed in the "Functional Programming" section since it's not related to functional programming in any way. So it should be located in the same category as "Faux Pas".

2. I would not delete Faux Pas, it's not like you forked a dead project but wrote another, arguably better, solution for the same issue.

robertvazan:
@akullpp You understand your list's policies best. I will resubmit the patch with these two issues addressed.

robertvazan:
Superseded by pull request #573.

Issue #525, opened 2017-05-04
Add (distributed) transactions section?


nmcl:
Was wondering if there'd be value in adding a section on transactions projects such as Bitronix and Narayana? Although local transactions are often a part of relational databases, these transactions projects can be used with a variety of data sources and at least one of them also supports forward compensation models, software transactional memory and RESTful transactions. If the idea is accepted I'll raise a pull request.

akullpp:
See 1023d1d85d593513030583c6cb0ed76182c12a8c

PR #572, opened 2017-06-18
Added distributed transaction section

nmcl:
As per https://github.com/akullpp/awesome-java/issues/525

There are a few older examples, such as Geronimo and JOTM, but these aren't being developed these days as far as I can tell so I didn't include them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,9 +337,9 @@ A curated list of awesome Java frameworks, libraries and software.
 
 *Distributed transactions provide a mechanism for ensuring consistency of data updates in the presence of concurrent access and partial failures.*
 
-* [Atomikos](https://www.atomikos.com) - Provides transactions for REST, SOA and microservices, also with support for JTA and XA.
-* [Bitronix](https://github.com/bitronix/btm) - A simple but complete implementation of the JTA 1.1 API..
-* [Narayana](http://narayana.io) - Provides support for traditional ACID and compensation transactions, also complies with JTA, JTS and other standards. (Previous called Arjuna Transactions).
+* [Atomikos](https://www.atomikos.com) - Provides transactions for REST, SOA and microservices with support for JTA and XA.
+* [Bitronix](https://github.com/bitronix/btm) - A simple but complete implementation of the JTA 1.1 API.
+* [Narayana](http://narayana.io) - Provides support for traditional ACID and compensation transactions, also complies with JTA, JTS and other standards.
 
 ## Distribution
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -332,6 +333,14 @@ A curated list of awesome Java frameworks, libraries and software.
 * [OpenTSDB](http://opentsdb.net) - Scalable and distributed time series database written on top of Apache HBase.
 * [OrientDB](https://orientdb.com/orientdb/) - Embeddable distributed database written on top of Hazelcast.
 
+## Distributed Transactions
+
+*Distributed transactions provide a mechanism for ensuring consistency of data updates in the presence of concurrent access and partial failures.*
+
+* [Atomikos](https://www.atomikos.com) - Provides transactions for REST, SOA and microservices, also with support for JTA and XA.
```
akullpp: Is Atomikos commercial? If so, please include the symbol.

akullpp:
Does it make sense to specify the type of transaction management with "distributed"? Of course it makes sense to use them in a distributed environment but is it a strict requirement? Maybe it would make more sense to go as abstract as possible first and then we specify the type as soon as the list gets larger.

README.md
nmcl: There's a company behind Atomikos but I believe there's a "free" download of the source (no free repo afaik). If the distinction is important let me know how best to proceed.

nmcl:
With the exception of Bitronix, the other two both offer distributed transaction support so it might be a small list of "local only" transaction managers.

nmcl:
One thought might be to categorise them by standards (e.g., JTS and JTA) they support, then perhaps other distribution mechanisms, e.g., REST/HTTP, and then perhaps models (ACID, extended, ...) Could get quite complex.

Moshe-Immerman:
Is it just me, but I am not sure I would call distributed transactions awesome? Isn't the whole trend and push behind microservices and the surrounding distributed tooling ecosystem about avoiding the need for coordination between systems and teams and allowing each service to own it's own data store?

I would think this would be more appropriate for an awesome-jee list

nmcl:
Transactions (distributed, local, ACID, extended) are pretty awesome and have existed before Java EE. They're also finding uses in microservices - check out various workshops and conferences in 2016 and 2017.

nmcl:
Sorry, but is there anything else I need to do here? If so, please let me know.

akullpp:
Fixed descriptions and the section of distributed databases which were actually deleted.

PR #573, opened 2017-06-19
Added NoException

robertvazan:
This is a refinement of pull request #571. NoException takes functional approach to exception handling. It allows throwing checked exceptions in functional interfaces and converts  all exceptions to Optional return to simplify fallback code.

NoException is a superset of functionality found in other similar libraries (listed on [NoException site](https://noexception.machinezoo.com/)). Arguably, NoException also has a cleaner API.

Disclosure: I am the author of NoException.


jbduncan:
Hi @robertvazan, sorry for leaving this for so long, and thanks for the PR!

I'm happy with it, so I'll merge it in on the assumption that @akullpp's happy as well.

PR #574, opened 2017-06-19
Add Jest - Elasticsearch REST client

heldev:
A nice common library supported by Spring Boot OOTB.  It's still a good choice because the official client is very low level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,9 +233,9 @@ A curated list of awesome Java frameworks, libraries and software.
 * [H2](http://h2database.com/) - Small SQL database notable for its in-memory functionality.
 * [HikariCP](https://github.com/brettwooldridge/HikariCP) - High-performance JDBC connection pool.
 * [JDBI](http://jdbi.org/) - Convenient abstraction of JDBC.
-* [Jedis](https://github.com/xetorthio/jedis) - A small client for interaction with Redis, with methods for commands.
-* [Jest](https://github.com/searchbox-io/Jest) - Elasticsearch REST client, supports JSON queries and also has own Java DSL.
-* [jetcd](https://github.com/justinsb/jetcd) - A client library for etcd.
+* [Jedis](https://github.com/xetorthio/jedis) - Small client for interaction with Redis, with methods for commands.
+* [Jest](https://github.com/searchbox-io/Jest) - Client for the Elasticsearch REST API.
+* [jetcd](https://github.com/justinsb/jetcd) - Client library for etcd.
 * [Jinq](https://github.com/my2iu/Jinq) - Typesafe database queries via symbolic execution of Java 8 Lambdas (on top of JPA or jOOQ). 
 * [jOOQ](http://www.jooq.org/) - Generates typesafe code based on SQL schema.
 * [Liquibase](http://www.liquibase.org/) - Database-independent library for tracking, managing and applying database schema changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -233,7 +233,8 @@ A curated list of awesome Java frameworks, libraries and software.
 * [H2](http://h2database.com/) - Small SQL database notable for its in-memory functionality.
 * [HikariCP](https://github.com/brettwooldridge/HikariCP) - High-performance JDBC connection pool.
 * [JDBI](http://jdbi.org/) - Convenient abstraction of JDBC.
-* [Jedis](https://github.com/xetorthio/jedis) - A small client for interaction with Redis, with methods for commands.
+* [Jedis](https://github.com/xetorthio/jedis) - A small client for interaction with redis, with methods for commands.
+* [Jest](https://github.com/searchbox-io/Jest) - Elasticsearch REST Client, supports JSON queries and also has own Java DSL.
```
jbduncan: Hi @heldev, would you kindly revert "redis" back to "Redis" please? AFAICT, "Redis" seems to be used more in the Redis documentation than "redis", even though the term "redis" is used in the website title header.

I've also spotted a couple of other minor spelling/grammatical errors that could be fixed, e.g.: "Elasticsearch" -> "ElasticSearch", and "Client" to "client".

heldev:
@jbduncan sorry, I changed that row by mistake. Client -> client change is done. I'm a bit surprised about "Elasticsearch" -> "ElasticSearch" because "Elasticsearch" is the official name (spring uses it too).

jbduncan:
> I'm a bit surprised about "Elasticsearch" -> "ElasticSearch" because "Elasticsearch" is the official name (spring uses it too).

Oh yes, you're right! I'm terribly sorry. I'm not entirely sure why I thought "ElasticSearch" was the correct spelling, but whatever the case, I didn't do my research. :P

tedyoung:
Don't feel bad, @jbduncan, even the Jest repo mixes it up quite a bit (and so does the referenced DeveloperWorks article!).

jbduncan:
Even though I just approved these changes, it's occurred to me that I don't know if **Databases** is the right section for this, considering that we already have a **Search** section which itself contains Elasticsearch.

@heldev @akullpp What do you think?

heldev:
I had doubts where to put it. 
**Search** section has "_Engines that index documents for search and analysis" description so it doesn't match well because Jest is just a client. So I've decided to use **Databases** mostly because of there are a couple of Redis-related items. But would be great if you suggest a better place.

jbduncan:
The way I see it, neither Jest nor Elasticsearch are really database-related, and I too have doubts now as to whether Jest belongs under **Search** since, as you pointed out, it's just a client for Elasticsearch.

So I suggest moving it to **Miscellaneous**. It's the section we use when something doesn't neatly fit into another existing section.

If we get another PR for something similar to Jest in the future, then that would be a good time to create a new section, e.g. **Search Clients**, and to expand the name of the **Search** section to **Search Engines** for clarity.

heldev:
I thought about **Miscellaneous** but it doesn't look like a section where people expect to see/search for an elastic-client. Do you think it's possible to categorize elastic as a [NoSQL-related](https://www.elastic.co/blog/found-elasticsearch-as-nosql) and leave Jest in "Database" together with all other NoSQL items (eXist, jetcd, Jedis ... ) ?

jbduncan:
@heldev Sorry for leaving this for so long. Looks like @akullpp agrees with your opinion on this, and I can see where you're coming from, so I'm happy with having them in **Databases** as well.

Issue #410, opened 2016-09-02
Ancients


akullpp:
See discussion in #94.

> The scope of awesome should probably not include outdated (e.g. Velocity) or highly disregarded frameworks (e.g. JSF)

How do we handle these frameworks?

jbduncan:
We could put a tiny legend (or equivalent section) towards the top of the README.

In my mind, it would have symbols like :star: and :x: (with brief descriptions saying that :star: means a project/framework is **awesome** and :x: means it's "ancient" or otherwise listed only for historical purposes and should be avoided in new code).

One would then label each project in the rest of the README with the appropriate symbol.

If using just 2 symbols is not sufficient to describe all the different types of **awesome** and "ancient"/undesirable/outdated projects, we could always add more to the legend.

Indeed, if :star: and :x: are not very good choices, we could go as far as considering other emojis.

jbduncan:
An alternative approach could be to append each project name with a single word describing it's level of awesomeness.

E.g. if we consider the **Build** section, it may become something like this:


## Build

*Tools which handle the build cycle and dependencies of an application.*

* [Apache Ant](http://ant.apache.org/) _(ancient)_ - Build process management with XML.
* [Apache Maven](http://maven.apache.org/) _(awesome)_ - Declarative build and dependency management which favors convention over configuration. It might be preferable to Apache Ant which uses a rather procedural approach and can be difficult to maintain.
* [Bazel](http://bazel.io) _(awesome)_ - Build tool from Google that builds code quickly and reliably.
* [Gradle](http://gradle.org/) _(awesome)_ - Incremental builds which are programmed via Groovy instead of declaring XML. Works well with Maven's dependency management.

jbduncan:
@akullpp, what are your thoughts regarding what I've suggested?

akullpp:
I would only mark specific libraries with an icon. Dividing the entries in two clusters and adding icons accordingly would make the list look bloated.

Also it is implied that every library in the list is some kind of 'awesome'. 😄

jbduncan:
Okay, your comment regarding marking specific libraries makes sense to me. :smile:

Apologies if I gave the impression that we should split the list into two clusters. What I had meant to say was that, somewhere in the list, maybe the top or the bottom, we should have a tiny 'legend' (in the Microsoft Excel sense, rather than legendary sense), and in this legend there would be e.g. two icons with corresponding descriptions:

- :star: - An awesome, shiny new project!
[...]

PR #577, opened 2017-06-25
refactor test category in multiple sub groups

Moshe-Immerman:


README.md
```diff
@@ -86,17 +86,15 @@ A curated list of awesome Java frameworks, libraries and software.
...
 * [Java Modeling Language (JML)](http://www.eecs.ucf.edu/~leavens/JML/) - Behavioral interface specification language useful for specifying the behavior of code modules. It combines Eiffel's Design by Contract approach with the model-based specification approach of the Larch family of interface specification languages, with some elements of the refinement calculus. Used by several other verification tools.
 * [JavaCC](https://javacc.org/) - Parser generator with syntactic lookahead.
 * [JavaServer Faces](https://javaserverfaces.java.net/) - Oracle's open-source implementation of the JSF standard, Mojarra.
 * [JavaServer Pages](https://jsp.java.net/) - Common templating for websites with custom tag libraries.
-* [JUnit](http://junit.org/) - Common testing framework.
 * [Launch4j](http://launch4j.sourceforge.net/) - Wraps JARs in lightweight and native Windows executables.
 * [Quartz](https://github.com/quartz-scheduler/quartz) - Open-source job scheduler library with Apache 2.0 license.
-* [TestNG](http://testng.org/) - Testing framework inspired by JUnit and NUnit, with different functionalities.
```
jbduncan: I agree with @heldev here. Once something has been declared "ancient", I think there needs to be a good reason to no longer declare it as such.

Consider putting JUnit and TestNG back in **Ancients**, renaming _JUnit_ to _JUnit 4_, and replacing the empty space left by removing "JUnit" from  **Testing Frameworks** with "JUnit 5" (which at this point deserves to be called _awesome_, methinks :wink:) and an appropriate description.

Moshe-Immerman: I think in this particular case, readability and usability of the list outweigh the ancient factor.  

i.e. If I am looking to take my testing to a new level, I would look for something in the same category as what I am already using (JUnit/TestNG). 

There is also an argument to be made to get rid of the ancients section entirely, by removing some of the items (e.g. JSF, JSP, Hudson, Glassfish, ant) and moving the rest to their own categories with perhaps a new "ancient" icon

heldev: Exactly, ancients section is not for "to take my testing to a new level", it's just for historic records. People are not supposed to use libraries from it.

jbduncan: Moving the entries in **Ancients** back into their respective non-ancient sections, but labelling them as "ancient" in some way, sounds compelling to me. I made a similar argument some time ago in https://github.com/akullpp/awesome-java/issues/410#issuecomment-244346612, so I wouldn't mind awesome-java going this way. However I feel @akullpp should have a say on this first before taking it further.

akullpp:
I reworked the whole PR and decided to provide more grained categories.

Also I didn't include the new `###`-categories in the table of content since the `id` generated by GitHub would require ugly names like `Testing Miscellaneous`, `Testing Frameworks` and so on.

PR #578, opened 2017-06-30
add zipkin, opentracing

yedward:


adriancole:
OpenTracing is not a tracing system, rather a java client interface used to
report traces to certain tracing systems (think slf4j-api or the api of
codahale metrics, but for traces). Not sure if this warrants a
clarification or a different section.

PR #585, opened 2017-07-16
Add Maven Wrapper - analogue of Gradle Wrapper.

heldev:
Pretty useful and common util in Java world.
Example: Spring Boot

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,6 +557,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [JPad](http://jpad.io/) - Snippet runner.
 * [Lanterna](https://github.com/mabe02/lanterna) - Easy console text-GUI library, similar to curses.
 * [LightAdmin](http://lightadmin.org/) - Pluggable CRUD UI library for rapid application development.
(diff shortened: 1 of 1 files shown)


jbduncan:
Hi @heldev, thanks for opening this PR, and sorry for not getting back to you sooner.

Whilst I agree that Maven Wrapper is useful and sufficiently "awesome", I don't know if awesome-java is the right place for it. Yes, we do have Modernizr listed in awesome-java, but I'm not entirely sure why it was accepted as I believe both Modernizr and Maven Wrapper would better fit in an awesome-maven list, if it exists.

jbduncan:
Hmm, I've just done a quick Google search, and I can't seem to find an existing awesome-maven repo anywhere, not even in [awesome-awesome](https://github.com/sindresorhus/awesome). So unless someone would be willing to curate an awesome-maven list, perhaps awesome-java _would_ be the right place for Maven Wrapper after all.

@akullpp Do you have any thoughts on this?

akullpp:
I agree that we should add it for now until someone is brave enough to create an awesome-maven list.

PR #595, opened 2017-08-09
Add Contract Testing section

eddumelendez:


akullpp:
Spring Cloud is the umbrella for Spring Cloud Contract which can be found on the projects page of Spring.

Pact is not Java-specific, however we can discuss to add pact-jvm.

However according to the contribution guidelines it should be added to the Testing section.

akullpp:
Added with 02a2cc5d41d825c7724f99ff8fd23e8de1abe692

PR #601, opened 2017-09-03
fix: default to SSL

Moshe-Immerman:


akullpp:
Please update the libraries changed in #600 e.g. Glassfish

akullpp:
Too much to fix. I'll write a small script to check for HTTPS

PR #603, opened 2017-09-06
Spring Cloud Contract

earayu:
* [Spring Cloud Contract](https://cloud.spring.io/spring-cloud-contract/) - Spring Cloud Contract is an umbrella project holding solutions that help users in successfully implementing the Consumer Driven Contracts approach.


akullpp:
We generally don't add every Spring project as there are dozens. Also it is part of the Spring Cloud project which can be found under the general entry linking to https://spring.io/projects

PR #605, opened 2017-09-11
Move RxJava to Ancients

heldev:
For new projects developers should use Reactor (it's an official recomendation from the author of RxJava https://twitter.com/akarnokd/status/774590596740685824 )


akullpp:
Difficult decision, I don't think it's time to delete it yet as it is the "official" implementation of Reactive Extensions.

PR #606, opened 2017-09-11
Move Joda-Time to Ancients

heldev:
From the official site: "Note that from Java SE 8 onwards, users are asked to migrate to java.time (JSR-310) - a core part of the JDK which replaces this project."


akullpp:
I will remove it, there's no need to keep it.

Sylvain-Delafoy:
I'd like to appeal this decision. There are still a lot of projects out there that would benefit from this library.

Maybe for the same reason RxJava is still recommanded for pre-java8 projects. Not everybody is on a green-field project and Joda-time fills a real need. (so real that it's addressed by Java it's most recent releases.


PS: I agree that nobody should use pre-java8 runtimes now (with the great retro-compatibility in Java8 and the obvious security issues). However, Most projects stay with the older JDKs.

jbduncan:
I agree with appealing this decision.

@Sylvain-Delafoy If you wouldn't mind opening a new issue for this, then that would probably be best to move this discussion forward, as it's surprisingly easy to lose track of new comments in closed issues.

PR #608, opened 2017-09-14
Adding XMLBeam library

dmitart:
Had to work with XML after a while and this library is really awesome, modern and helpful.


akullpp:
See contribution guidelines 

> If an entry doesn't fit any of the pre-existing specialised sections, it should go under Miscellaneous.

> If two or more entries in Miscellaneous are in the same domain, then they can be moved to a new specialised section.

akullpp:
Added with c289a280a2978dc27c3ae51b5a9c46674b9f6771

PR #609, opened 2017-09-23
add beanmother

keepcosmos:
Please consider adding Beanmother lib :)


keepcosmos:
@akullpp Could you check the project? Should it be more popular?

PR #610, opened 2017-09-24
Remove Early-access releases of JDK 9

heldev:


akullpp:
I'm a bit confused, we removed JDK 9 already yet the PR didn't update. Please comment if I'm missing something.

PR #611, opened 2017-09-26
Add Unirest Java to HTTP Clients

Clivern:
Please consider adding [Unirest Java](http://unirest.io/java.html) to HTTP Clients. It is a great package and well maintained.


akullpp:
Doesn't seem to be maintained at all, look at there GH. Lot's of issues compared to other clients and a fluent API isn't really that unique.

PR #618, opened 2017-10-03
Move RxJava to Ancients again

heldev:
@akullpp I'm a bit confused :). I see that you merged the request and, at the same time, I see your comment and [Change Spring URL to projects](https://github.com/akullpp/awesome-java/commit/5c05e03ba2f504d537c8e5b23d875eeaa7c2902b) commit that effectively reverts it. Is it an accidental change?

PS anyway, I understand your concern. RxJava has a solid ecosystem but at the same time, it's hard to keep it on the list because even the author says to use a different technology if you can.  It seems like an exact match to Ancients section (a good technology but you should not start new projects with it).
Additionally, it can be really confusing to people who are using the list because RxJava has more starts, more mentions,  dedicated awesome list and, at the same time, the author's tweet by far is not the first link in the google search result for "RxJava vs reactor". 

What do you think ?


akullpp:
Sorry, wasn't my intention. Good catch!

PR #621, opened 2017-10-15
Added keywhiz under secuirty section.

sureshg:
Keywhiz is a system for managing and distributing secrets, developed by square.


akullpp:
Being a tool written in Java there could be arguments if it is suitable. I think it is and I will also add Vault.

PR #623, opened 2017-10-17
Cactoos: object-oriented Java primitives

ordepdev:


heldev:
Hi @ordepdev ,
Would it be possible to have some description why this project is awesome ?
1. Similar to #625 
2. "We're still in a very early alpha version, the API may and will change frequently."

akullpp:
Added under the premise of the Contribution guidelines section (c) and (d).

PR #625, opened 2017-10-21
Add Takes (https://github.com/yegor256/takes)

filfreire:
[Takes](https://github.com/yegor256/takes) is a object-oriented and immutable Java7 web development framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,6 +856,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [PrimeFaces](http://primefaces.org/) - JSF framework with both free and commercial/support versions and frontend components.
 * [Ratpack](https://ratpack.io/) - Set of libraries that facilitate fast, efficient, evolvable and well-tested HTTP applications.
 * [Spring Boot](http://projects.spring.io/spring-boot/) - Microframework that simplifies the development of new Spring applications.
+* [Takes](https://github.com/yegor256/takes) - Object-oriented and immutable Java 7 web development framework.
 * [Vaadin](https://vaadin.com/home) - Event-driven framework build on top of GWT. Uses server-side architecture with Ajax on the client side.
 
 # Resources
--- a/README.md
+++ b/README.md
@@ -856,7 +856,7 @@ A curated list of awesome Java frameworks, libraries and software.
 * [PrimeFaces](http://primefaces.org/) - JSF framework with both free and commercial/support versions and frontend components.
 * [Ratpack](https://ratpack.io/) - Set of libraries that facilitate fast, efficient, evolvable and well-tested HTTP applications.
 * [Spring Boot](http://projects.spring.io/spring-boot/) - Microframework that simplifies the development of new Spring applications.
-* [Takes](https://github.com/yegor256/takes) - Object-oriented and immutable Java 7 web development framework.
+* [Takes](https://github.com/yegor256/takes) - Object-oriented and immutable web framework.
 * [Vaadin](https://vaadin.com/home) - Event-driven framework build on top of GWT. Uses server-side architecture with Ajax on the client side.
 
 # Resources


heldev:
I really hate to say this but is there any example why this specific project is awesome?
I'm asking this question because usually, that guy a "public person" and produces extremely poor "products" if you look "under the hood". If it's a conference talk then people just leave the room, if it's a project then it's a just messy bunch of libraries that he puts together in the dirtiest way, and if it's a podcast interview he can't provide a good technical explanation of his point of view...

yegor256:
@heldev here is more about that guy: http://www.yegor256.com/testimonials.html

heldev:
I know all that stuff, but would be great to have some good technical explanations why this framework should be in this list

yegor256:
@heldev can you provide a good technical explanation why not?

heldev:
@yegor256 Because awesome lists don't work that way (see [CONTRIBUTING](https://github.com/akullpp/awesome-java/blob/master/CONTRIBUTING.md)). My argument pointed that the project source is suspicious and we should be a bit more careful with that PR.

akullpp:
I've thought about this for quite a long time and first of all I suggest that you should generally separate the author from his work if it's not inherently tied to views that violate GitHub's terms.

I looked at the code and its usage and personally I wouldn't use and wouldn't recommend it since it's too simplistic (not "simple") - only speaking subjectively since I can't point to a single objective principle.

However, it is in active development, has some popularity and a unique approach. So I say, if someone doesn't have valid technical arguments, I will add it at the end of the week.

Moshe-Immerman:
Perhaps we should make it  *Opinionated object-oriented and immutable web framework.*

heldev:
> generally separate the author from his work

In this case, it's difficult because views and approaches of the author and his company are one of the reasons why we should not expect any good projects from there. It's like a product company that is used to produce the cheapest products with the lowest quality, nobody buys something important from that brand. It's the same case, the author and his company already have a "public brand".

> I wouldn't use and wouldn't recommend it

This repository is some kind of modeled list of good projects. People use it to find something for own/prod projects. I don't see any reason why someone can even consider this framework.

> has some popularity

Probably because the author pushes it hard among newcomers, he even has tutorials and a webinar based on the framework

> it is in active development

 It's another illusion, author's projects use the same approach to keep "active development" status. Usually, there are tons of "random" commits that modify author's email, formatting back and forth, years and licenses, fix "quality control issues", which are produced by author's tools qlice and xcop that are maintained using the same approaches. There is no difference between 1.7.1 and 1.7 for library users. The last "normal" code change was after 1.3.1 where the author reinvented Supplier interface and introduced it everywhere.

Here is the diagram of compile dependencies for the project.

 Jcabi and xembly are author's projects. All of them are wrappers or naive implementations of existing projects. Here is an example from jcabi-immutable, project description:
image

>There are a few classes that are truly immutable (not like those pseudo immutables from Guava and Apache Commons):

Naive O(N) implementation of Map::get in ArrayMap:
https://github.com/jcabi/jcabi-immutable/blob/98234fcc9320b282b2208e0f543c3238ec961bed/src/main/java/com/jcabi/immutable/ArrayMap.java#L243
```java
    @Override
    public V get(final Object key) {
        V value = null;
        for (final Map.Entry<K, V> entry : this.entries) {
            if (entry.getKey().equals(key)) {
                value = entry.getValue();
                break;
            }
        }
        return value;
    }
```

The framework does nothing and it's already slower than even Spring http://www.yegor256.com/2015/03/22/takes-java-web-framework.html#comment-1954760366

There are other technical comments which point to the meaningless of the project on the same page. 

Probably, I can't gather more but taking into account that web frameworks section is already big I hope I will be enough.

amihaiemil:
@heldev Ironically, if you, who didn't know anything about Takes, were able to compile such a detailed analysis, it means it's a quite a well designed project, right?

The diagram is rather simple, which is OK -- the less dependencies, the better. How will this diagram look for Spring or other frameworks? 

And if the author has their own quality control tools, and fixes the project to respect those tools, what is wrong in that? 

Lastly, the jcabi components are quite cool and very productive. Try using jcabi-http and jcabi-github -- undeniably some bad motherfuckers, I build all my project with them and have 0 regrets.

heldev:
@amihaiemil, sorry, but it's hard to discuss this topic with you because you are working for the author (or at least you are member of jcabi and teamed organizations)

jbduncan:
I've just had a look through the commit log for Takes, and I'm drawn to the apparently huge number of commits which just make relatively small changes to config files, tests and other files here and there. Furthermore, I had a quick look at the commit log for xembly, and that project seems to tell a similar story.

Therefore I'd personally vote to close this PR.

amihaiemil:
@heldev yes, I was working on the jcabi projects (and I think I made a few commits on Takes? Not sure) -- but they are really good, I'm not just saying for support. :)

funivan:
I really hate to say this but is there any example why Spring is awesome from the technical side? I think this is unreal to provide technical specification for project awesomeness. 

My awesome rule: if I use it, and do not want to uninstall it - then project is awesome. 
I'm not working in Yegor's company but I use takes and cactoos and this libraries are awesome for me.

heldev:
@funivan you are a Cactoos contributor

Now it's really hard to say who has an "independent" point of view because all the company workers has git hub accounts and there are a lot of them. But I see that the author clearly organized a thumb up and PR support flash mob. So I'll probably stop participating in the thread here.
But this is one the reasons why the source of the project has triggered my initial reaction - the company prefers blind dirty public marketing over technical excellence.

amihaiemil:
@heldev at a minimum, Takes is innovative through its approach, there is no other framework with such ideas in mind. And just because you do not like the idea, or the author as a person, it doesn't mean it's not eligible for this list :D 

And which company are you talking about? Takes is under Yegor's name and if I remember well, Teamed stopped paying for OSS contributions about the time Takes was created. People who contributed really liked the idea, nobody did it for money :)

terciodemelo:
I don't like the project. That being said, it is indeed innovative; the amount of (IMHO) unpleasant commits is a downside, but every project have downsides.

Because of it's unique approach to the domain, it certainly deserves to be in a public list for the public to judge, see, experiment and contribute if they like it. Open source should not be as toxic as this thread.

The fact we don't like it doesn't mean it's not a contribution (a good one) to the community.

I don't have any connection to the framework's author, nor have I contributed with it; in fact I don't contribute much with anything from anyone... for now... :)
I see myself as impartial/independent enough, I say this PR should be merged

PR #633, opened 2017-11-17
Add sshj - SSHv2 library for Java

heldev:
A popular client (1000+ stars) not as powerful as MINA but it has a straightforward API. Here is some [comparison with other popular options](https://medium.com/@ldclakmal/comparison-of-commons-vfs-sshj-and-jsch-libraries-for-sftp-support-cd5a0db2fbce).


heldev:
The library doesn't match the description of the section ("Libraries for building network servers") but it seems like a good place taking into account the section name and surrounding libraries.

PR #634, opened 2017-11-18
adding Commons CSV library to CSV group

emanuelbatista:
Commons CSV - https://commons.apache.org/proper/commons-csv/


jbduncan:
Thanks for the PR @emanuelbatista, but I highly doubt it will be accepted because @akullpp removed it from awesome-java recently (https://github.com/akullpp/awesome-java/commit/7fc5e8a5dfcf7827239012a41c67ec527bf08471) because we already have Apache Commons as an entry, in a similar way to how we already have Spring as an entry and so disallow Spring subprojects and extensions. :)

emanuelbatista:
@jbduncan I check your comment, I think you are right.

PR #650, opened 2018-01-13
Add Javalin web framework

tipsy:


akullpp:
Please provide a detailed description to help users see the unique points.

tipsy:
@akullpp provide it where? I made the description to the same level of detail as the other entries in the list 🤔

PR #662, opened 2018-03-14
Include fastest Java JSON library to the list

zapov:
https://github.com/eishay/jvm-serializers/wiki
https://github.com/kostya/benchmarks
https://github.com/fabienrenaud/java-json-benchmark


zapov:
btw, Jsoniter is on this list, but I dont think copying DSL-JSON architecture + code should be considered an awesome behavior.

PR #663, opened 2018-03-25
Adding Plummb.com - This is a resiliency framework for Java application

ankitjaininfo:
Plummb is a resiliency framework. Once added as a dependency and initialized, it automatically tracks exceptions, timeouts, logging aspects, etc. All these can be managed via a dashboard. Once up and running, no further coding needed for fault-tolerance.


akullpp:
Is it closed-source? Indicate the free plan and pricing (via the icon we use for other libraries). Also I don't see the advantages compared to other products.

akullpp:
Closed due to inactivity.

PR #665, opened 2018-04-05
Added java-express

CliffordJ:
https://github.com/Simonwep/java-express


jbduncan:
I'm a bit concerned about including java-express, given that it's licensed under GPL v3. If people were to use it, would it not mean that they couldn't use it in closed-source commercial applications?

jbduncan:
Having said this, we already include [urnlib](https://github.com/slub/urnlib), which is also licensed under GPL v3, so maybe my fear is unfounded?

CliffordJ:
Yeah, but it looks like java-express is still relatively young. The licence was also changed in the last commit (from MIT to GPLv3) maybe they change their opinion and publish it again under MIT.

jbduncan:
@CliffordJ Ah, okay.

Regardless of whether java-express goes back to using the MIT license, I don't have a strong preference personally as to whether this library should be included in awesome-java or not.

@akullpp WDYT?

PR #667, opened 2018-04-09
Add zerocell excel library to document processing section

zikani03:
Added zerocell excel library to document processing section


akullpp:
I'm not convinced of the library. Only reading Excel files is a quite narrow scope. The bonus is the annotation processing I think. You should make this clear in the description.

PR #671, opened 2018-05-03
edited jmonkey game engine link

NavyaZaveri:
The old link is dead, added the new one.


akullpp:
The link is not dead, the certificate *.github.com is just untrusted. They should probably use Cloudflare.

PR #681, opened 2018-07-19
Add Apache Ant to list of build tools

alanktwong:
Lots of Java projects still use Apache Ant for their builds.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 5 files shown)


heldev:
It was here before but was removed with all 'Ancient' projects
https://github.com/akullpp/awesome-java/commit/072859d9597bee6bac5f96f2517b9dd19b8541c5

akullpp:
Lot's of closed discussion about this one, just use the search.

PR #689, opened 2018-09-03
add jasync-sql -  Java async database driver for MySQL and PostgreSQL written in Kotlin.

oshai:
https://github.com/jasync-sql/jasync-sql
It is fairly new, but haven't seen other such drivers except for its deprecated predecessor: https://github.com/mauricio/postgresql-async


oshai:
Awesome, thanks!

PR #691, opened 2018-10-03
openliberty link added

jerrak0s:
openliberty link added


akullpp:
`Servers` would be the correct section, wouldn't it? It seems to be a lightweight version of the WebSphere AS, right?

jerrak0s:
@akullpp yes `servers` section would be more appropriate. I will edit the pr

PR #668, opened 2018-04-12
Add ASCII/Console helpers

FTOH:

PR #693, opened 2018-10-03
Added Entry asciitable - Library to draw tables in ASCII manner in console output

zugzug90:
I hesitated in which category should I add this lib so any suggestions are welcome.


akullpp:
I want to merge this and #668 but I will probably need to do it by hand.

Issue #698, opened 2018-10-17
Source for TLS Connection


IAmPramod:
Found a great source to understand how TLS Connection works. We should add this in the networking section.

https://tls.ulfheim.net/

IAmPramod:
Will raise a PR for this shortly.

akullpp:
How is this related to the Java programming language?

PR #699, opened 2018-10-18
[#698] Add TLS Connection Illustration

IAmPramod:
- Add a informative guide to understand how a TLS Connection works.
- In the demonstration, a client has connected to a server, negotiated a TLS 1.2 session, sent "ping", received "pong", and then terminated the session.


akullpp:
Please see the discussion at #698

PR #709, opened 2018-11-22
Open JDK distribution list

osigida:
a list of Open JDK Distributions by different vendors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ A curated list of awesome Java frameworks, libraries and software.
   - [JSON Processing](#json-processing)
   - [JSON](#json)
   - [JVM and JDK](#jvm-and-jdk)
+    - [Open JDK Distributions](#open-jdk-distributions)
   - [Logging](#logging)
   - [Machine Learning](#machine-learning)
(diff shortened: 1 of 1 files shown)


akullpp:
Please don't create a single subsection, this PR would be fine if you add the entries to the JDK section, don't you think?

jbduncan:
[Azul's Zulu OpenJDK](https://www.azul.com/downloads/zulu/) should also be mentioned in this PR, but other than that and and @akullpp's suggestion, this LGTM. :)

osigida:
> Please don't create a single subsection, this PR would be fine if you add the entries to the JDK section, don't you think?

@akullpp, Open SDK Distributions are not something new or different to OpenSDK by Oracle, it is just a packaging by different vendors, to avoid future licence issues with Oracle. So it supposed to be the same as Open SDK.
That is why I added it as a sub list to SDK list.

However, I feel it is important to keep this way or another.
Please let me know if you still insist to merge it to SDK list, and I'll update the PR.

akullpp:
I fixed the PR, there has been duplicates and missing formats.

PR #719, opened 2019-02-02
message queue  opensource qmq

Devonmusa:
https://github.com/qunarcorp/qmq

Selling point
this is  mature  message queue  middleware,qunar company used it from 2012 ，different design with kafka。
1.it non partition，easy extend consumers
2.support  transaction message
3.seconds level delay message
4.monitor dashboard  
5.message trace support opentrace protocol
6.monitor metrics support prometheus  or other

So I think that this library fulfills aspects (c) and (d).

Thank you!


akullpp:
Unfortunately the documentation is only in Chinese. Is there an English translation somewhere?

PR #720, opened 2019-02-04
added Cache2k

svenruppert:


cruftex:
https://cache2k.org

Some selling points:
- fast/fastest and memory efficient in process cache implementation as an alternative for EHCache
- smallest, but complete JCache compliant cache implementation available
- works also with Spring, Hibernate and on Android
- the only library that has comprehensive support for Nulls and Exceptions

PR #733, opened 2019-03-27
Update README.md to add brpc-java for networking

wenweihu86:


vdiravka:
@wenweihu86 Please edit the description of your PR.
Also I believe that `brpc-java` should be translated into English for easy use of this library.

akullpp:
Although I might consider adding your library, it is imperative that it has a documentation so that we provide the possiblity that the global developer community can use it and not only a subset.

PR #736, opened 2019-04-04
add brpc-java to networking section

wenweihu86:
brpc-java is java implementation of brpc (https://github.com/apache/incubator-brpc).
brpc-java is maintained by baidu.com, and widely used in baidu.com. 
the main feature are:
1. support multi-protocol, and can be easily extend to arbitrary protocol.
2. extremely high performance.
3. the core jar and spring support is separated.
4. support multi naming type, and can be easily extend to arbitrary naming service.
....


akullpp:
Please read the contribution guidelines, we are a global community and therefore require documentation in English.

PR #737, opened 2019-04-13
Add dregex

marianobarrios:
Dregex is a regular expression engine that uses deterministic finite automata. It supports some Perl-style features and yet retains linear matching time, and also offers set operations. AFAIK, there is nothing similar in Java.

The library is written in Scala, but it has a fully Java-frienly API, and documentation.


akullpp:
Would be better suited for awesome-scala.

PR #740, opened 2019-04-29
[XerJ.CommandStack and XerJ.EventStack](https://github.com/XerProjects/xerj.commandstack and https://github.com/XerProjects/xerj.eventstack) - Small, easy-to-use command/event handling libraries that has first class support for dependency injection pattern and frameworks.

jeyjeyemem:
* XerJ.CommandStack
This project composes of components for implementing the command handling parts of the CQRS pattern. 
XerJ.Command enforces a one-to-one mapping between commands and command handlers as described by the CQRS pattern. It only allows registering a single command handler for a command and throws an exception if done otherwise.

* XerJ.EventStack
This project composes of components for implementing the event handling parts of the DDD/CQRS pattern.
XerJ.EventStack supports dispatching event messages to one or many registered event handlers. This is useful for capturing events that happened in the system and reacting by doing some side-effect operations.

This library was built with simplicity, modularity and pluggability in mind. 
Support for various dependency injection libraries are provided by the library for easy configuration/detection of command handlers within your projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ A curated list of awesome Java frameworks, libraries and software.
   - [Compiler-compiler](#compiler-compiler)
   - [Configuration](#configuration)
   - [Constraint Satisfaction Problem Solver](#constraint-satisfaction-problem-solver)
+  - [CQRS](#cqrs)
   - [CSV](#csv)
   - [Data structures](#data-structures)
   - [Database](#database)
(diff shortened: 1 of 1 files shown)


akullpp:
I'm not sold on the idea creating a new section for CQRS if there are other opinions we can discuss it in the issues.

jeyjeyemem:
@akullpp at which section should I put it in?

PR #746, opened 2019-05-21
add serenity bdd

globalworming:
worked with this framework for half a year now, it's great. community and development is active.
unique selling points: implements the screenplay pattern, two-way jira integration ...

- Write about it's unique selling point compared to other projects.  [x]
- If it's a commercial project, then mark it as such, e.g. `[Title ![c]](URL)`. [x]
- Ensure that you provide concise and informative descriptions. [x]
- Do not use a description like "A library/project/tool/framework for JSON processing in Java" since all of this is implied. [x]
- Finish the description with a dot. [x]
- Try to order it alphabetically. [x]


heldev:
Non https link seems a bit suspicious (especially when the https version of the same url shows a certificate error). Why not https://www.thucydides.info ?

globalworming:
oh, sorry, you are totally right. didn't realize..
changed it

globalworming:
are there any other issues you want to address before putting serenity on the list?

PR #756, opened 2019-07-22
Add Armeria http client/server

zhenik:
Added [Armeria](https://github.com/line/armeria) library. 

It's a unique project because of next features:
* client and server side
* HTTP/2
* Integration with gRPC and Thrift
* Asynchronous and reactive

It's not a commercial project, Apache License 2.0.


trustin:
Doesn't Armeria also fall into 'Microservice' category?

zhenik:
@trustin yes, could be

PR #764, opened 2019-10-02
Add Spring Boot to Microservices section

phxql:
Hi,

i added Spring Boot to the Microservices section. Spring Boot is one of the most well-known frameworks for microservices on the JVM.


akullpp:
The Spring project is already added, please feel free to add a section like Apache but no single Spring projects.

PR #766, opened 2019-10-02
avoid redirects if possible, use https

Ghost93:


heldev:
When I go to http://www.choco-solver.org/ my browser changes the URL to http://www.choco-solver.org (no slash at the end). Do you think we could put the version without that slash ?

Ghost93:
@heldev Done at 03d693f

PR #786, opened 2019-10-04
Add Beagle under Search

dainiusjocas:
Beagle is a detector of interesting things in text. Its intended use is in-stream search applications. Suppose you need to monitor a stream of text documents such as web crawl results, chat messages, or corporate documents in order to identify keywords, phrases, regexes, and complex search queries of interest. With Beagle you can quickly be up and running with such a system, allowing you to focus on productively monitoring your documents.

Beagle is based on the Lucene monitor library which is based on Luwak.

Compared to Lucene Monitor it has a data-driven API which makes using the functionality a lot easier.


akullpp:
I don't think this belongs to the search category but to NLP as it provides other stuff too. Also would you please link to the GitHub page and not to the wiki? Thanks.

PR #788, opened 2019-10-08
#Added EasyXLS

almeidaah:


akullpp:
Commercial product with a Java library isn't it? I don't think it's that awesome compared to the alternatives. Other opinions?

heldev:
Agree, looks very enterprisy, would be nice to see an example when a dev would choose it over other common libs

PR #791, opened 2019-10-16
Add GC4S

hlfernandez:
Add GC4S, a library of reusable and customizable GUI components for Java Swing.


akullpp:
Any other opinions? I think Swing isn't the way for GUI development in modern Java.

thisisleobro:
I think Swing should not be ignored.

dariodariodario:
Indeed, intelllij is in swing

PR #796, opened 2019-10-31
Add SmartDec SAST

ATSDec:
Add information about **SmartDec** to **Code Analysis** section.

SmartDec Scanner is a SAST tool which is capable of identifying vulnerabilities and backdoors (undocumented features). Its distinctive feature is the ability to analyze not only source code but also executables without debug info (i.e. binaries) and to return much better results than when using DAST. The analyzer can test apps written in more than 30 programming languages or that have been compiled into an executable file with one of nine supported extensions, including those for Google Android, Apple iOS, and Apple macOS. SmartDec Scanner also can be easily integrated into a Secure SDLC process.
Moreover, SmartDec Scanner does run a vast variety of checks, automating vulnerability discovery as well as detecting generic errors and discrepancies with recommended code writing practices. Such checks obviously does help Java developers better understand language semantics as well as properties of their code base and ensure guideline acceptance for it.


akullpp:
Anyone has an opinion? I've never used it, so I can't say much.

akullpp:
Doesn't have a free tier (only a requested trial), which clashes with the new contribution guidelines.

PR #808, opened 2019-12-10
Reintroduce Hamcrest Matchers

drothmaler:
Reintroduce Hamcrest matcher library, as it has been reactivated and updated for Java 9++ about a year ago. So I think it is still/again a good alternative to AssertJ/Truth.


akullpp:
Sounds good.

PR #809, opened 2019-12-10
Update Google Truth Link and Description

drothmaler:
* Link to the project page, instead of the source repo
* Mention that it provides "fluent" assertions - as there is a big difference in usage between fluent assertions and simple (or nested) assertion methods


akullpp:
Thank you for the clarification!

PR #821, opened 2020-01-31
Move Armeria to Microservice which is more inline with its features

anuraaga:
I guess a client is an aspect of most microservice frameworks anyways (or maybe it should be listed in both sections?). But Armeria's main features are for developing microservices, not clients so I think this is more appropriate.

/cc @trustin @adriancole


adriancole:
agreed microservice frameworks include client support but are far more than that.

trustin:
Thanks, @anuraaga! :+1:

PR #825, opened 2020-02-26
add more batching and scheduler frameworks

mabujaber:
add more batching and scheduler frameworks


akullpp:
Please add a more helpful description and don't just use the short tag lines from GitHub. Also the entries have a certain format which should be consistent, e.g. final dot we also hava a linter if you would like to verify it.

akullpp:
Merged with 1ebb6e4

PR #834, opened 2020-02-28
Add QueryStream.

archiecobbs:
QueryStream provides a `Stream`-like API for building JPA queries.

For example:

```java
    public double getAverageSalaryOfDirectReports(Employee manager) {
        return qb.stream(Employee.class)
          .filter(e -> qb.equal(e, manager))
          .flatMap(Employee_.directReports)
          .mapToDouble(Employee_.salary)
          .average()
          .value();
    }
```

It's probably not going to meet your criteria (too new and not widespread), but thought it was worth a try due to its novel approach. Althought open source, it is well tested and is currently being using in mission-critical production commercial applications.

> (a) widely recommended, regardless of personal opinion
> (b) highly discussed in the community due to its innovative nature

Scores low on these. It's very new and not at all widespread.

> (c) absolutely unique in its approach and function

Is similar to Jinq, but does not attempt to analyze bytecode and maps more clearly and directly onto JPA functionality.

> (d) a niche product that fills a gap

Yes, the gap is that normal JPA programming is a PITA.

You won't hurt my feelings if you reject this PR. I just thought I'd ask.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,7 @@ _Everything that simplifies interactions with the database._
 - [MariaDB4j](https://github.com/vorburger/MariaDB4j) - Launcher for MariaDB that requires no installation or external dependencies.
 - [Modality](https://github.com/arkanovicz/modality) - Lightweight ORM with database reverse engineering features.
 - [Presto](https://prestosql.io) - Distributed SQL query engine for big data.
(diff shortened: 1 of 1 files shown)


akullpp:
Looks fine too me. :+1:

PR #837, opened 2020-03-14
Add library for tail call optimization

kag0:
javac doesn't have tail call optimization. I'm not aware of any other libraries out there that offer it. 
Since java 8 and progression of more functional features (esp upcoming pattern matching), recursive methods will become increasingly important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ _Augmentation of the development process at a fundamental level._
 - [Manifold](https://github.com/manifold-systems/manifold) - Re-energizes Java with powerful features like type-safe metaprogramming, structural typing and extension methods.
 - [NoException](https://noexception.machinezoo.com) - Allows checked exceptions in functional interfaces and converts exceptions to Optional return.
 - [SneakyThrow](https://github.com/rainerhahnekamp/sneakythrow) - Ignores checked exceptions without bytecode manipulation. Can also be used inside Java 8 stream operations.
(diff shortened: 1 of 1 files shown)


akullpp:
Need more opinions on this, not sure about the code quality.

kag0:
As far as code quality, it's about 70 lines. If you want to check it out you might be able to get a sense of the code quality without investing too much time.

heldev:
Usually I prefer to "inline" such kind of projects (sneaky throw is a similar Java case).
I mean I think it's a safer option to spend some time reading the code and reimplement it artistically in your service (even though you might need it in every repo). And when I say safer it's not only about security but also about elegance and persuading/onboarding your team...

akullpp:
It's 70 lines code, you would need to add a specific repository to get it... Also there's the pattern of Subramaniamin in his 2014 book Functional Programming in Java. I guess we could add it and see how it goes.

heldev:
I didn’t hear name Subramaniamin before, but I’ve heard how people refer to this workaround as “trampoline” and there are big mature libraries that implement it (cyclops family is on the list, though never worked with it)
https://github.com/aol/cyclops-integration/wiki/Trampoline-:-Stackless-Recursion-for-Java-8

@kag0, @akullpp just take a look if it changes your mind

kag0:
It looks like the separate dependency-free module doesn't include all the functionality of the library being proposed, but cyclops-core seems complete and well implemented.

PR #842, opened 2020-04-19
adding sql-to-mongo-db-query-converter

vincentrussell:
sql-to-mongo-db-query-converter helps you build quieres for MongoDb based on Queries provided in SQL.


akullpp:
Lacks quality.

PR #845, opened 2020-04-23
docs: add imgix-java to imagery libraries

frederickfogerty:
This PR adds imgix-php to the list of awesome components. This is an awesome library because it makes images rendered by it automatically more performant, and responsive! The repository has been around for 5 years, well-documented, commercially-supported.


heldev:
It’s a client library provided by the CDN service
So it seems more like a commercial CDN entry than a Java specific project

PR #850, opened 2020-05-16
Add findbugs to Code Analysis secction

rohit909:
FindBugs, a program which uses static analysis to look for bugs in Java code.  It is based on the concept of bug patterns. A bug pattern is a code idiom that is often an error.

Find-bugs: http://findbugs.sourceforge.net/


heldev:
It was on the list but [was replaced by its successor SpotBugs](https://github.com/akullpp/awesome-java/commit/7e5fc969325981218f19455ecfb306ca9c5e6a77)

PR #864, opened 2020-07-19
Added SSLContext Kickstart

Hakky54:
SSLContext Kickstart provides a high-level factory class for configuring a http client to communicate over SSL/TLS for one way authentication or two way authentication. It has the ability to easily load any amount of key and trust material. It is lightweight as it has almost none transitive dependencies.


akullpp:
You should consider renaming it, the naming is not very attractive  ;-)

Hakky54:
Hi @akullpp thank you for merging! Which part of the naming is not attractive?

PR #868, opened 2020-08-20
Added olivomarco/lc4j

olivomarco:
I am adding my github project olivomarco/lc4j which is new to Github but I had hosted that on my website for a very long time. 

I developed this library back in 2005 porting and adapting an existing project from Perl, and over time I know it has been incorporated in some projects. I think it is a good addition to the list because the other NLP libraries do not offer to my knowledge language determination of a snippet of text.


akullpp:
We normally don't add programs and to use it as a library you'll need to compile it from scratch. Best to add it a repository so it's easier for interested people to add it. Also you are severely lacking tests. ;-)

PR #872, opened 2020-08-27
Add ActiveSerializer

valerialistratova:
ActiveSerializer is one of the fastest Java serializers in the world. At the same time, it is extremely powerful and features full support of Java subclasses, collections (including Maps), and also specialized collections like HPPC. Works directly with Java classes via annotations. No additional layers of intermediate DTO classes, which significantly speeds up the code and makes it easy-to-use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,7 @@ _Libraries related to handling date and time._
 
 _Libraries that help to realize the [Inversion of Control](https://en.wikipedia.org/wiki/Inversion_of_control) paradigm._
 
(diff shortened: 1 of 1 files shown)


heldev:
Looks like you have a whole brand new platform there :)
@akullpp do you think it's a good idea to have a section for them under Platforms ? Overall these libraries seem to be components of a bigger project (similar to Micronaut). Are there any maturity "requirements"?

akullpp:

PR #873, opened 2020-08-28
Add ActiveInject

valerialistratova:
ActiveInject is a lightweight and powerful dependency injection library with ultimate performance and no third-party dependencies. It is multi-thread friendly, feature-rich, and can boast with lighting-fast start-up time and runtime, significantly outperforming Spring DI and Guice according to the benchmarks (https://github.com/activej/activej/tree/master/benchmarks/inject/src/main/java/io/activej/inject). In ActiveInject all reflection is entirely optional and abstracted out in a separate module. Also ActiveInject is fully transparent for introspection of the dependency graph


akullpp:

PR #874, opened 2020-08-28
Add ActiveCodegen

valerialistratova:
ActiveCodegen is an efficient Java code generator with streamlined and simple API. It is built on top of ObjectWeb ASM and has zero-overhead performance.


akullpp:

PR #875, opened 2020-08-28
Add ActiveSpecializer

valerialistratova:
A unique technology that automagically speeds up code by rewriting the bytecode in runtime. Unlike traditional compiler optimization techniques, it uses runtime information of classes instances.


akullpp:

valerialistratova:
> I think it's fine to only add ActiveJ and not all subprojects.

Ok, thanks. Can I create a pull request to add ActiveJ to the **Platforms** section then?

PR #879, opened 2020-08-31
Add guava

davidlanouette:
Although Guava is included in the Utilities section, the caching section should mention it, because the guava caching library is used widely, and by itself.


heldev:
Wasn't it superseded by [Caffeine](https://github.com/ben-manes/caffeine) ?

davidlanouette:
Hmm.  I didn't know about this.  From my brief reading, it sounds like Caffeine is not a "replacement" for Guava Cache, but very much inspired by it, and a likely successor.

Never the less, I'm fine with closing this PR.

Thanks.

PR #880, opened 2020-09-01
Add Spock

Charlew:


dariodariodario:
does spock have a java API?

jmsalcido:
it's groovy, IMO it should be part of this repository with a disclaimer, since tests are written in groovy for the JVM.

st-h:
It's a broadly accepted way to test your java code. And it's awesome imho (just in case anyone cares 😂)
However, it seems to have it's history with this repository already https://github.com/akullpp/awesome-java/search?q=spock&type=Commits

jmsalcido:
@st-h I care :) thanks for the info about already discussed. Probly we should not add this... again.

PR #886, opened 2020-09-03
Add myself to the People section of this awesome repo

neomatrix369:
Add my twitter handle and description to the People section


neomatrix369:
There is a mix of both Java and AI, although I agree there is an inclination a bit more on AI - what do you suggest in this case?

Issue #885, opened 2020-09-03
How can I add a link that spans through many resources related to Java/JVM?


neomatrix369:
I have been gathering a lot of Java-related resources related to AI/ML/DL here:
- https://github.com/neomatrix369/awesome-ai-ml-dl/blob/master/details/java-jvm.md#java

and about GraalVM, Wasm, and related tech here

- https://github.com/neomatrix369/awesome-graal

So these are not libraries or frameworks or packages but resources that are a collection of a whole lot of things

Would it be okay to add it under Documentation or Miscellaneous or Awesome lists or all of them? Please guide me, as I'll be creating a PR and would be happy to discuss this on it.

akullpp:
See PRs.

PR #888, opened 2020-09-03
Adding awesome-ai-ml-dl: java link to the Awesome section

neomatrix369:
Adding the Java aspect of the Awesome-AI-ML-DL github repo to the Awesome lists section

Related to issue #885

### DIFF
--- a/README.md
+++ b/README.md
@@ -1087,6 +1087,7 @@ _Frameworks that handle the communication between the layers of a web applicatio
 
 _Awesome lists related to the Java & JVM ecosystem._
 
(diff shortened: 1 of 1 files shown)


akullpp:
I'm not sure about that list since there is very little Java-specific content or am I missing something?

neomatrix369:
> I'm not sure about that list since there is very little Java-specific content or am I missing something?

Yes, true, the Java-specific content is less compared to that of the Python and other offerings which also reflects the reality of things in this space where Java and AI/ML/DL overlap, although this is fast changing. 

The purpose of sharing this link is to bring in this small list closer to the Java developer/community and hoping this grows and expands.

My personal efforts is also grow this area in resources and hence this list.

PR #911, opened 2020-11-27
Added scene2d.ui to the GUI section.

NathanSweet:
There are very few options in the GUI section for Java and IMO scene2d.ui deserves more attention.

scene2d.ui is part of [libgdx](https://libgdx.com) and is built on top of [scene2d](https://github.com/libgdx/libgdx/wiki/Scene2d), which is a 2D scene graph. scene2d provides a solid foundation for organizing scene graph nodes, drawing, receiving events, and applying actions over time. scene2d.ui adds on top of that a layout mechanism, a collection of UI widgets, and a way of skinning widgets. Layout specifically is crucial for building UIs and one of scene2d.ui's key innovations is 1) doing it in a sane way and 2) providing an easy to use layout mechanism sufficient for almost all layout needs: [Table](https://github.com/libgdx/libgdx/wiki/Table).

scene2d.ui is libgdx-based, so runs on desktop/web/mobile. [Spine](http://esotericsoftware.com) is an example of a sophisticated desktop app. [Gloomhaven Helper](http://esotericsoftware.com/gloomhaven-helper) is an example for desktop/[web](http://esotericsoftware.com/gloomhaven-helper-web)/mobile ([Android](https://play.google.com/store/apps/details?id=com.esotericsoftware.gloomhavenhelper) and [iOS](https://apps.apple.com/us/app/gloomhaven-helper/id1456538503)).

scene2d.ui is part of libgdx, so is Apache 2 FOSS. It is production ready and stable, as it has been around ~10 years, is used in established apps such as those listed above, and is widely used by libgdx users.

NB: I am a libgdx author/cofounder, but I do think scene2d.ui fits in your list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,7 @@ _Libraries to create modern graphical user interfaces._
 
 - [JavaFX](https://wiki.openjdk.java.net/display/OpenJFX/Main) - Successor of Swing.
 - [Scene Builder](https://gluonhq.com/products/scene-builder/) - Visual layout tool for JavaFX applications.
+- [scene2d.ui](https://github.com/libgdx/libgdx/wiki/Scene2d.ui) - OpenGL-based GUI toolkit for desktop/web/mobile. 
 - [SWT](https://www.eclipse.org/swt/) - Graphical widget toolkit.
 
 ### High Performance


akullpp:
I need to think about this one since it is not standalone and we already have libgdx

NathanSweet:
True. The poor GUI section is so thin, maybe an exception can be made to draw attention to a viable GUI solution. Either way you decide, thanks for considering it!

PR #912, opened 2020-11-28
Created computer vision section. Moved JavaCV and added BoofCV

lessthanoptimal:
- Created Computer Vision section since there are now two libraries
- Added BoofCV
  *  Highly likely to be the most popular pure Java general purpose computer vision library
- FYI OpenCV is C++ and basically dominates the field in general
  * JavaCV (probably the most popular Java CV library) is a wrapper on top of OpenCV, as mentioned in this list
- Deeplearning4j could probably be added to this section since it is used for image processing too, but is a ML library at its core.

Disclaimer I'm the author of BoofCV, but feel free to double check with [Google](https://www.google.com/search?q=java+computer+vision&oq=java+computer+vision). In general, Computer Vision in Java is a bit of a niche. Going to add one more library in another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ A curated list of awesome Java frameworks, libraries and software.
   - [Code Coverage](#code-coverage)
   - [Code Generators](#code-generators)
   - [Compiler-compiler](#compiler-compiler)
+  - [Computer Vision](#computer-vision)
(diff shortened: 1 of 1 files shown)


akullpp:
Merged with  6ba7315

PR #913, opened 2020-11-28
Added ImageJ

lessthanoptimal:
- This is the second PR after the one which created the Computer Vision section
- Adds ImageJ which is a popular tool in medical image processing with a large community
- Its omission is a bit glaring actually

(I have no affiliation with this project)


shathor:
It's missing the dash between link and description.

lessthanoptimal:
fixed and squashed

PR #920, opened 2021-01-27
Technical Analysis Libraries added

MrABKhan:
Technical Analysis Libraries : 
                     Since the list was missing TA libraries which are used in day-trading and other financial firms for analysis.


shathor:
There's a *Financial* section. I think it makes more sense to have them there. For those unfamiliar the term "Technical Analysis" is not associated with the Finance industry.

akullpp:
Added ta4j to financial.

PR #923, opened 2021-02-19
Added Gatekeeper

Hakky54:
### Link
[GitHub - Gatekeeper](https://github.com/Hakky54/gatekeeper)

### Short description
Gatekeeper is a library which guards your publicly accessible internal implementations.

### Description
Library maintainers have a hard time to keep internal implementations internal and prevent it to be accessible by library users. The library maintainer cannot easily change internal implementations as it may lead into breaking changes for the library users. A straight forward method for library maintainers would be making classes package protected and accessible through public api's so-called Service classes. Encapsulation internal api's and exposing it through service classes is the preferred way but cannot be achieved in some use cases. Restrictions by Java Modules will not work if the end-user is not using java modules. Gatekeeper will ensure that these internal implementations remain internal by validating the caller. You as a library developer can choose who is allowed to call your classes.


akullpp:
Not sure about this and will keep it open for a while. It's a super simple (basically 3 LOC) programmatic way to check the caller's class at runtime. If you want to check the architecture constraints ArchUnit is a better way to do it.

Hakky54:
Yes, I also wasn't sure if it was worth adding it, that was the reason why I closed it earlier. Still not sure but I thought maybe it could be useful for some developers

PR #925, opened 2021-02-21
Adding commons-networking

liran2000:
Adding commons-networking

Features:

- SSE (Server-sent Events) client

Simple SSE client almost without any dependencies, did not found anything similar online, and not here in this list.

It is used in developments in a similar version.

As it is new and similar implementation not found, it is currently still cannot be widely recommended.

Note: I am the author of this SSE client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,7 +717,7 @@ _Libraries that specialize in processing text._
 
 _Libraries for building network servers._
 
-- [Commons-networking](https://github.com/CiscoSE/commons-networking/tree/main/commons-networking#commons-networking) - SSE (Server-sent Events) client.
+- [Commons-networking](https://github.com/CiscoSE/commons-networking) - Client for server-sent events (SSE).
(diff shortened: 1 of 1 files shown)


akullpp:
I'll keep this open for a while. I think it's not ideal to have the SSE client and the GNMI utils in the same project and probably should be split up to two separate modules.

liran2000:
You are correct. Both are on separate modules already, just the repository
is common. If wanted, each can be linked to its specific module

בתאריך יום ב׳, 1 במרץ 2021, 10:07, מאת Andreas Kull ‏<
notifications@github.com>:

liran2000:
By your suggestion, I updated the PR changes to contain only the SSE client, and the link is referring only to the specific module of SSE client.

liran2000:
@akullpp can you merge it ? any concern ?

This is updated just for SSE Client module.

PR #928, opened 2021-03-11
Update readme.md - added virus scanning library

laurajbouchard:


shathor:
What specifically has this to do with Java?

laurajbouchard:
The virus scanning library offers free APIs that allow you to scan files and cloud storage for viruses and other malware in Java

akullpp:
I don't have a good feeling about this. An API can be called from any client.

PR #934, opened 2021-05-03
Added "DevProjects" under "Miscellaneous"

jenniferfu0811:
I recently found DevProjects, a community with all kinds of free coding projects, and added it to Miscellaneous because I've been using the website's projects to practice Java. I'm quite new to Java, so I focus on the easier ones. I personally find the solutions and discussions helpful and think building Java projects is a fun way to learn Java.

If you think it might be a good resource for the community, I personally think it'd be a great addition!


akullpp:
There's a website section, but I don't think we'll add it without more requests. I guess we could also add something like LeetCode if we add DevProjects which is somewhat strange.

PR #937, opened 2021-05-22
Add workflow engines section 

### LINKED ISSUE #910: Distributed workflow systems
I keep bumping into tech (both OS and commercial) that is supposed to replace manually coded sagas and similar in distributed systems. Today it was https://github.com/temporalio/temporal / https://github.com/uber/cadence . I didn't have a chance to try them, but I see in real-life that people write their variations of them.  I use this list as a starting point of many investigations and if/when I need to make a choice I would appreciate having such a category. Ideally with a meaningful search-friendly description

Do we want to include them?
Or is there anyone who tried rolling them out at an organization and can tell if it's a good/bad idea?

heldev:
Long overdue PR to resolve #910
Conductor from Netflix and Zeebe from Camunda didn't make the list because they use yaml/json DSLs or visual builders (though they have Java clients).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ A curated list of awesome Java frameworks, libraries and software.
   - [Version Managers](#version-managers)
   - [Web Crawling](#web-crawling)
   - [Web Frameworks](#web-frameworks)
+  - [Workflow Orchestration Engines](#workflow-orchestration-engines)
 - [Resources](#resources)
   - [Awesome Lists](#awesome-lists)
   - [Communities](#communities)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1113,6 +1114,11 @@ _Frameworks that handle the communication between the layers of a web applicatio
 - [Takes](https://github.com/yegor256/takes) - Opinionated web framework which is built around the concepts of True Object-Oriented Programming and immutability.
 - [Vaadin](https://vaadin.com) - Event-driven framework that uses standard web components. Server-side architecture with Ajax on the client side.
 
+### Workflow Orchestration Engines
+
+- [Cadence](https://cadenceworkflow.io) - Stateful code platform from Uber.
+- [Temporal](https://temporal.io) - Microservice orchestration platform, forked from Cadence but gRPC based.
+
```
heldev, line 1121: The descriptions and the section naming are the tough part would be glad if somebody suggested something better

PR #938, opened 2021-05-24
Added Certificate Ripper

Hakky54:
[Certificate ripper](https://github.com/Hakky54/certificate-ripper) is a java based CLI tool to easily extract the full chain of any server/website either in pem format or human-readable format. The end user can inspect any sub fields and details easily on the command line. Next to that it can also save the certificates in a p12 container. 

## Demo
alt text

It has an executable jar, also a separate executable for Windows, Linux and Mac OS X is available so the end-user can run it without java.


akullpp:
Nice tool, but it's not related to the Java programming language which this list is about.

Hakky54:
Thank you for your feedback!

PR #940, opened 2021-06-10
added oauth and microservice framework in readme file

nayekabhishek1:
- Added new OAuth section in the list.
- Added Keycloak in OAuth framework section.
- Added Grails in Microservices section.


akullpp:
Keycloak already exists, Grails is on a rather steep downfall.

PR #962, opened 2021-11-05
chore: add questdb to database category

pswu11:
- Add QuestDB to the database category

> If it's a commercial project, then mark it as such, e.g. `[Title ![c]](URL)`.

Hi, QuestDB at the current stage is not commercialized, but in the future perhaps will be. Should I mark it? 
Thanks a lot for maintaining this nice list!


akullpp:
What's the relation to the Java programming language other than it was used to build the DB?

pswu11:
Hi @akullpp QuestDB has an embedded API and so it can also be used as a Java library.  https://questdb.io/docs/reference/api/java-embedded

PR #963, opened 2021-11-09
Add Jollyday to Date Time category

SuperPat45:
The Jollyday  API determines the holidays for a given year, country/name and eventually state/region. The holiday data is stored in XML files (one for each country) and will be read from the classpath. You can provide your own holiday calendar XML file or use any of the provided ones. Currently there are 63 countries supported like the following: United States, most european countries, Russia, India, Australia. Besides those there will be more special calendars like ...


shathor:
The repository has moved here: https://github.com/svendiedrichsen/jollyday

PR #967, opened 2021-12-23
Update README.md

KirillZi:
One more database schema migration tool for your list. Similar to Flyway but does not depend on JVM . Currently focused on PostgreSQL.
Samples: https://github.com/DBinvent/rdbm


shathor:
But what is the connection to Java then?

akullpp:
With Flyway can integrate it in your Java code, that's why it's on the list. Maybe your suggestion is better suited for the awesome-database list?

PR #977, opened 2022-04-11
Updated Useful Link

sandeepmishratech:
Have Added Scaler Topics Under Websites. Scaler Topics Has programming tutorials on Java, DSA, Python etc.


akullpp:
Not strictly related to the Java programming language and it's not clear what the unique selling point is of that particular site.

PR #986, opened 2022-07-08
added webvium browser

mrepol742:


shathor:
Note that this is an Android App... There is no Java API. I don't see any direct relation to Java. It is not open source (?), its licensing is generally unclear.
Also typo "injeting".

mrepol742:
> Note that this is an Android App... There is no Java API. I don't see any direct relation to Java. It is not open source (?), its licensing is generally unclear. Also typo "injeting".

Hi,

The javascript code was functioning using webkit javascript interface in Android Java https://developer.android.com/reference/android/webkit/JavascriptInterface
It was been closed source.
The app is a freeware.

PR #988, opened 2022-07-24
Add Javalin

ManuInDenWolken:
Javalin is a common-used, minimal purpose web framwork for creating simple and lightweight backends.


shathor:
Javalin was removed in PR: https://github.com/akullpp/awesome-java/pull/858 as it is not written in Java.

ManuInDenWolken:
See https://github.com/akullpp/awesome-java/pull/992#issuecomment-1195180437

PR #990, opened 2022-07-24
Add Xtend

ManuInDenWolken:
Xtend is an older Java dialect, often used in connection with Xtext for compiler generation.


shathor:
It's another programming language.

ManuInDenWolken:
It's a dialect. Based on Java. And *for* Java.

shathor:
You could argue the same for Groovy. Most Java code is even syntactically valid Groovy code. Xtend is pretty much like any other JVM language.

ManuInDenWolken:
You have to differ between a dialect and a completely new language. A dialect builds only upon it's base and ensures inevitably interoperability. In this case, it would be only connected with a Java language context, especially in combination with Xtext. But here and in this case, I would like to read further arguments and explanations from further participants.

PR #1012, opened 2022-10-03
Add Error Prone Support to Code Analysis

japborst:
Error Prone Support is the first library that open sources Refaster templates on top of https://github.com/google/error-prone. Additionally, there's been significant work done to make it super simple to rewrite Java code at scale.


japborst:
Not a fork of Error Prone. Builds on top of.

PR #1014, opened 2022-10-07
Adds ZK framework

maiconandsilva:
Every word used to describe the ZK framework were taken from https://www.zkoss.org/product/zk and https://www.zkoss.org.

Github repository: https://github.com/zkoss/zk
License: https://github.com/zkoss/zk#license


maiconandsilva:
It's "enterprise-y" in some sense, but as a developer with about 2 years using this framework developing internal systems for a government institution together with a small team using the free open source version, I'd say it covers most needs and when it doesn't it's possible to extend it, customize it or maybe integrate another framework like Spring (which it offers support) into it.

Pros:
- It's extensible through its own component and template declaration language or using JSP taglibs;
- It's customizable: You get nice (I think 4 or 5) looking themes for free and you can customize them yourself, although you get a lot more themes (23) with the paid version;
- Its free component library covers most needs. The paid components are extra and most of the times can be replaced even with the help of jsp tags
 
Cons:
- Community is a little smaller compared to other frameworks like Primefaces

Couldn't think of any more arguments to build the case against this framework for now but it certainly has more cons than just one.

PR #1016, opened 2022-10-23
Add a new framework in Other section in Platform category.

m0ver:
Add a new framework in Other section in Platform category.


m0ver:
> I don't see much of documentation or an indepth explanation of the use case, maybe you can elaborate a bit and provide documentation.

Is this also can be a reason to be rejected? :(

PR #1039, opened 2023-02-09
Add Instancio

armandino:
Instancio generates fully-populated objects based on class. It also provides a JUnit 5 extension for reproducing failed tests.


armandino:
@akullpp, thank you!

PR #1041, opened 2023-02-13
Add Bencher

epompeii:
https://github.com/bencherdev/bencher


epompeii:
> Relation to the Java PL?

@akullpp , there's an adapter for [JMH](http://openjdk.java.net/projects/code-tools/jmh/), which is an already listed tool.
This allows Java developers to store the results of their JMH benchmarks, compare those results, and catch performance regression.

Issue #1040, opened 2023-02-13
Add Hippo4j


mageeric:
Can we add https://github.com/opengoofy/hippo4j/blob/develop/README-EN.md under the monitoring section?

@akullpp If you are ok with the above-mentioned project kindly let me know, I will add it to the list.

PR #1042, opened 2023-02-20
Add Hippo4j (#1040)

magestacks:
Dynamic and observable thread pool framework.


magestacks:
English document address:[README-EN](https://github.com/opengoofy/hippo4j/blob/develop/README-EN.md)

magestacks:
thanks

PR #1065, opened 2023-08-30
Adding Embedchain JS framework to NLP section

aryankhanna475:
- Embedchain JS is a framework to help create your own personalised ChatGPT like  bot with just few lines of code in less than 10 minutes. Repo: https://github.com/embedchain/embedchainjs
- Its completely opensource since day 1
- Its python version is already trending with 4.1K+ stars (https://github.com/embedchain/embedchain)
- Description has been kept short and its added in the alphabetical order in the section.


hoppermann:
I think a Javascript or even Python framework, as useful as it is, has no place in a list for Java.

PR #1071, opened 2023-10-23
Rename Monitoring to Observability/Monitoring

jonatan-ivanov:
Observability is a wider term and there are items under monitoring that are not strictly monitoring tools but they might fit better under the umbrella of Observability.


jonatan-ivanov:
Hi @akullpp,
Did you close this accidentally?
If not, could you please explain why you closed it.

akullpp:
It might add more confusion, I changed the description so that experts can
find it. Maybe we can change it later, let me think more about it.

PR #1107, opened 2024-07-16
Add various manifold references, add new section for GraphQL

rsmckinney:


akullpp:
We have the [Platform](https://github.com/akullpp/awesome-java?tab=readme-ov-file#platform) section for these kind of projects, please add it there.

rsmckinney:
@akullpp Thanks, but I think there is a misunderstanding; this is not a platform. If you re-examine the change, you'll see that each link is a separate distinct reference to a separate library. For instance, the SQL reference is a completely separate library from the JSON reference; they operate independently. Perhaps what is confusing is that they are all published under the same "manifold" domain.

Arturo171 (approved):

akullpp:
Sorry, I still don't get it. Aren't these items just extensions/plugins/features to manifold? None of those can be used standalone, right?

PR #1135, opened 2025-03-09
Polish description for MyBatis-Plus

izeye:
This PR polishes the description for the MyBatis-Plus a bit.


andrzejsydor (approved):

akullpp:
Great resource, but we have an international community, so an English translation would be required!

PR #1150, opened 2025-06-30
Add WebForms Core technology

elanatframework:
WebForms Core is a new technology for handling HTML tags from the server. It is a two-way protocol between the [WebForms class](https://github.com/webforms-core/Web_forms_classes) on the server side and the [web-forms.js](https://github.com/webforms-core/Web_forms) library on the client side, where processing is done on the client side and the server sends Action Control commands to WebFormsJS.
WebForms Core is a radical technology that eliminates the need to use front-end frameworks. This technology is available in most server-side programming languages.
The following link shows how to use WebForms Core technology in Java with an example:
https://github.com/webforms-core/Web_forms_classes/tree/elanat_framework/java


akullpp:
Not sure about this one, the association to the Java PL is a bit loose.

PR #1156, opened 2025-08-17
Adding JSpecify Package-Info Generator

bcaillard:
Since `JSpecify` is becoming the standard for unifying all nullity annotations (it is now used by `Spring Framework` v7, for example), I created this `Maven` plugin to avoid having to manually place an annotation on all classes or having to create a `package-info.java `file each time.


vega0707:
你好，邮件已收到，谢谢

PR #1163, opened 2025-11-06
Add Spring Data Dynamic Query

tdilber:
Added a new entry for Spring Data Dynamic Query to the list.

It is my more than 6 years old project. I started to develop it in a hobby project and extract it and I made a compelete project. 

### Project abilities: 
- Using Serializable objects to build Dynamic Queries. It means you can take this objects from an api or you can do hardcoding. 
- All queries actually created from Spring Data apis (in JPA using Criteria API). There is no cheapy tricks. So It is safe for sql injection etc. 
- Production-Ready. (I have been using about 3 years.)
- Well-tested. Each database type has between 75-80 test method. 
- Project still has up-to-date for both Sring boot 2 and spring boot 3 dependencies at the same time for latest version. (because i need still sb2 dependencies :) )
- Dependencies are Pushed in maven central. 8 depedencies (sb2 and sb3) pushed 2x core, 2x jpa, 2x mongodb, 2x elasticsearch 
- Same JpaRepository interface and same models (placed on core module) are used for all 3 database type Spring Boot Jpa(SQL), Spring Boot MongoDb, Spring Boot Elasticsearch.
- Easy to open-close with Spring standart Enable Annotations. 
- Supported Statements: Select, distinct, Where, Order By, Paging

### Statements Abilities: 

**FROM Statement:** 
- There is no actual from statement, your base entity selected from which repository using (like normal JpaRepository). 
- Automatic easy join. inner join 2 table using only dot(.) , left join two table using (<). You can join multiple like this (roles<roleAuthorizations<authorization.name).
- This joins operation are not operated by my project. Criteria Api knows which tables are joined looking for (`@ManyToMany`, `@OneToMany` ... etc) native Jpa Joining annotations. 
- If you need join in only Select or Where or Order By statement, system will automaticly join. There is no limitation. 

**SELECT Statement:** 
- Type 1:  Annotation based Easy Projection. Nested Object supported, record supported, one annotation auto joining tables. 
- Type 2: (My Older projection) Non annotated objects auto projection. (Using reflection) instantiate with empty constructor. after that try to fill fields, Ignore If missing fields   
- Type 3: Spring Data style default Entity return. All Jpa abilities working on this return type. Fetch Eager, Lazy etc.

**WHERE Statement:** 
- 4 string operator like `CONTAIN`, 6 comparator operator like `EQUAL` or `GREATER_THAN`, null check operator supporting when using `WHERE` statement. Multi value `EQUAL` supporting like `IN`. 
- `AND`, `OR` and `SCOPE` operations supporting.
- Auto joining supporting 

**ORDER BY Statement:** 
- Multi order by condition supported. 

**PAGING Statement:** 
- Page Size and page count supported. 
- All methods have Spring Data standart returning `Page<>` object and `List<>` object. 

### Additional Features: 
- Query DSL like query building. (repo.queryBuilder().select(..).where(..).orderBy(..).page(1, 10).getResultAsPage()). 
- Argument Resolver for Dynamic Query. Has different EnableAnnotation. Easy to take DynamicQuery from GetMappings. Example: `/api/users?key0=name&operation0=EQUAL&values0=John&key1=name&operation1 ...`
- Custom Converter for deserialize received data to entity data type. 
- Update with Annotated Object in JPA. Save in one transaction for joined (multi table) Object data. 

**Note:** Handwritten text. I don't want to use AI in this texts. Because I think it will make more boring. But I still want to share summary of chatGpt :)  

### ChatGpt Summary: 

Spring Data Dynamic Query is a unified, production-ready library that dramatically simplifies dynamic query creation across three major data stores: JPA, MongoDB and Elasticsearch. It eliminates the verbosity and boilerplate typical of runtime queries by offering a single, consistent API layer. Unique features include full support for dynamic JOINs (inner and left joins in JPA, nested object joins in Elasticsearch), an expressive criteria engine supporting nine operators (e.g., CONTAIN, GREATER_THAN, PARENTHES for scoped logic), and powerful projection models via @JdqModel and @JdqField—even enabling Java record-based DTOs with nested fields. The library seamlessly handles SELECT, DISTINCT, ORDER BY, paging and projection in one go, and—most notably—brings type-safe, minimal-code update operations with dot-notation paths into joined relations (for JPA only). Since it reuses Spring Data’s underlying Criteria APIs, it remains secure and performant while giving you a high-level, declarative query mechanism. In short: if you’re tired of writing countless repository methods or hand-rolling query builders for every new filter or projection, this library delivers one clean solution that works across multiple stores and covers the full spectrum from WHERE through projection to update. It’s not just a convenience—it’s a compelling upgrade to how you work with Spring Data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,7 @@ _Everything that simplifies interactions with the database._
 - [Redisson](https://github.com/redisson/redisson) - Allows for distributed and scalable data structures on top of a Redis server.
 - [requery](https://github.com/requery/requery) - Modern, lightweight but powerful object mapping and SQL generator. Easily map to or create databases, or perform queries and updates from any Java-using platform.
 - [Speedment](https://github.com/speedment/speedment) - Database access library that utilizes Java 8's Stream API for querying.
+- [Spring Data Dynamic Query](https://github.com/tdilber/spring-data-dynamic-query) - Unified dynamic query interface for Spring Data JPA, MongoDB, and Elasticsearch, enabling advanced JOIN(s), OR logic, scoped conditions, powerful projections and advanced features with zero boilerplate.
 - [Spring Data JPA MongoDB Expressions](https://github.com/mhewedy/spring-data-jpa-mongodb-expressions) - Allows you to use MongoDB query language to query your relational database.
 - [Trino](https://trino.io) - Distributed SQL query engine for big data.
 - [Vibur DBCP](https://www.vibur.org) - JDBC connection pool library with advanced performance monitoring capabilities.


andrzejsydor (approved):

tdilber:
Hi,I believe my suggestion will add value to both your repo and my repo. I would be happy for my repo to be included in your collection.

Thank you for your interest.

andrzejsydor (approved):

PR #1245, opened 2026-05-14
Add Javalin to Web Frameworks

fauzan171:
Added Javalin - lightweight REST framework for Java and Kotlin.


andrzejsydor (approved):

akullpp:
Thanks for the suggestion. Javalin is already listed under Web Frameworks, so this would be a duplicate entry.

